PR #2, opened 2021-01-14
run validator and display json output using http response

lilian-hunt:


aannie-y (commented):
lgtm

webapp/src/main/webapp/index.html
```diff
@@ -9,15 +9,6 @@
...
-      <h2>About</h2>
-      <p>The General Transit Feed Specification (GTFS) is an open data standard. It enables transit agencies to share their data describing their transit service in a standard format that can be used by rider-facing software applications. </p>
-      <h2>How to use</h2>
-      <p>Upload a compressed .zip file including all of the individual files for your transit service.</p>
-      <form method="post" action="fileupload" enctype="multipart/form-data">
-        <input type="file" name="file" id="file" accept=".zip" data-multiple-caption="{count} files selected"/>
-        <br>
-        <button class="btn btn-outline-primary" type="submit">Upload</button>
-      </form>
     </div>
```
jupi3200, line 12: Why is this file changing?

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +83,56 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+    final long startNanos = System.nanoTime();
+
+    // Input.
+    feedLoader.setNumThreads(numThreads);
+    NoticeContainer noticeContainer = new NoticeContainer();
+    GtfsFeedContainer feedContainer;
+    GtfsInput gtfsInput;
+
+    try {
+      gtfsInput = GtfsInput.createFromPath(Paths.get(filePath));
```
jupi3200: Can this be tested?

webapp/src/main/webapp/index.html
lilian-hunt: I removed it from here to include it in another PR request. I wasn't sure how otherwise to create the new PR without redoing all my branches

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: I will figure out how to write the tests now

webapp/src/main/webapp/index.html
jupi3200: Please be careful when merging the PRs.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
jupi3200: Thanks

lilian-hunt: I just created 2 simple tests that ensure it reads file and generates the correct json

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +84,56 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+      Files.write(Paths.get("/tmp/", "report.json"),
+          noticeContainer.exportJson().getBytes(StandardCharsets.UTF_8));
+    } catch (IOException exception) {
+      exception.printStackTrace();
+    }
+
+    final long endNanos = System.nanoTime();
+    System.out.println(
+        String.format("Validation took %.3f seconds", (endNanos - startNanos) / 1e9));
+    System.out.println("Table totals: " + feedContainer.tableTotals());
```
jupi3200: Do we need to keep it?

lilian-hunt: Sorry what exactly are you referring to?

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +84,56 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
             // Save the file on disk
             item.write(storeFile);
 
-            // TO DO: Call validator here 
+            // TO DO: get feed name and thread specification from form
+            String validatorOutputJson = runValidator(filePath, 1, "nl-openov");
```
jupi3200: What does 'nl-openov' mean?

In this case the code might be more readable if you used something like _runValidator(filePath, /* numThreads= */ 1, /* feedNameString= */ "nl-openov");_
or better assign the values to variables to help make the code clear.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +84,56 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+        feedLoader.loadAndValidate(gtfsInput, feedName, validatorLoader, noticeContainer);
+
+    // Write the output to the temporary file: what happens when multiple files are written?
+    // Need a unique name ... or just write it to reponse and display it as HTML?
+    try {
+      Files.write(Paths.get("/tmp/", "report.json"),
+          noticeContainer.exportJson().getBytes(StandardCharsets.UTF_8));
+    } catch (IOException exception) {
+      exception.printStackTrace();
+    }
```
jupi3200: Why do we need to write the report to a file?

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +84,56 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+    // Input.
+    feedLoader.setNumThreads(numThreads);
+    NoticeContainer noticeContainer = new NoticeContainer();
+    GtfsFeedContainer feedContainer;
+    GtfsInput gtfsInput;
+
+    try {
+      gtfsInput = GtfsInput.createFromPath(Paths.get(filePath));
+    } catch (IOException e) {
+      e.printStackTrace();
```
jupi3200: Use java logger?

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: it is an example feed name, the feed name has to be in the for ab-abc where ab is a country code, I have the TO DO there to get the country code from the user in the form, ok will create variables for the thread and feedname

webapp/src/test/java/com/google/sps/servlets/RunValidatorTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mobilitydata.gtfsvalidator.input.GtfsInput;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsFeedLoader;
+import org.mobilitydata.gtfsvalidator.validator.ValidatorLoader;
+
+@RunWith(JUnit4.class)
+public class RunValidatorTest {
+  @Test
+  public void testSimpleGtfsFeed() {
```
jupi3200, line 27: It might be easier to test the method if it returned Json instead of a string.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: because the gtfs validator function reads the input from a file

webapp/src/test/java/com/google/sps/servlets/RunValidatorTest.java
```diff
@@ -0,0 +1,55 @@
...
+    Assert.assertEquals(json, expected);
```
jupi3200: https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

Can you possibly think of a more generic expectations that would fail if the verifier executed at all but would not be dependent on specific error detection?

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
jupi3200: The System.out.println (It looks like some debug output that is not needed.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
jupi3200: Yes, using a named variable would solve the issue.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +84,56 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
             // Save the file on disk
             item.write(storeFile);
 
-            // TO DO: Call validator here 
+            // TO DO: get feed name and thread specification from form
```
jupi3200: I don't think the client should determine the number of threads we use to execute the validator (as this is implementation detail of the service).
I would suggest using a named constant of 1.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
jupi3200: I am not referring to the input zip file. I am referring to the report file (report.json).

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: Ok yes, sorry should be removed. Maybe it would be useful to put in the logs. This is from the original CLI output in the GTFS main function. I have just made a checklist of things to review before submitting pull requests including remove print statements so I fix it all before the next pull request

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: Yes, that is what I was considering in the comment above. But I think you're right that we don't need to do this anymore

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: Using logger now

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: Ok sure, created a final static variable. The old CLI allowed the user to specify themselves, but they were running it locally on their machine so that makes sense

webapp/src/test/java/com/google/sps/servlets/RunValidatorTest.java
lilian-hunt: The mobility data function creates the json in the function and the calls gson.toJson() and returns this. So do you think it would be best to call from fromJson on the return value? Otherwise it would involve changing the mobility data function
Screen Shot 2021-01-18 at 10 24 18 am

lilian-hunt: I am now returning the notice container

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -76,13 +90,49 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
             // Save the file on disk
             item.write(storeFile);
 
-            // TO DO: Call validator here 
+            // TO DO: get feed name from the user form
+            String feedName = "nl-openov";
+            NoticeContainer validatorNotices = runValidator(filePath, NUM_THREADS, feedName);
```
jupi3200: As the number of threads is fixed, do we need to define this variable to the runValidator method?

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -37,10 +47,14 @@
   private static final int MAX_FILE_SIZE = 1024 * 1024 * 40; // 40MB
   private static final int MAX_REQUEST_SIZE = 1024 * 1024 * 50; // 50MB
 
+  // Log any issues
+  private static final Logger LOGGER = Logger.getLogger(UploadServlet.class.getName());
```
jupi3200, line 50: s/LOGGER/logger/

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: True we don't, will remove this from the constructor.

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: Is this a link to a google resource? Because I do not have access to it (403 error)

webapp/src/test/java/com/google/sps/servlets/RunValidatorTest.java
lilian-hunt: I am now just checking that it identifies the error e.g. that this is contained in the notice "\"code\":\"invalid_row_length\",\"totalNotices\":13" with google's truth library

webapp/src/main/java/com/google/sps/servlets/UploadServlet.java
lilian-hunt: Ok will rename LOGGER to logger

PR #4, opened 2021-01-14
create html page to upload file

lilian-hunt:


aannie-y (commented):
lgtm

webapp/src/main/webapp/index.html
```diff
@@ -9,6 +9,15 @@
   <body>
     <div id="content">
       <h1><a href="index.html">GTFS Validator</a></h1>
+      <h2>About</h2>
+      <p>The General Transit Feed Specification (GTFS) is an open data standard. It enables transit agencies to share their data describing their transit service in a standard format that can be used by rider-facing software applications. </p>
+      <h2>How to use</h2>
+      <p>Upload a compressed .zip file including all of the individual files for your transit service.</p>
+      <form method="post" action="fileupload" enctype="multipart/form-data">
+        <input type="file" name="file" id="file" accept=".zip" data-multiple-caption="{count} files selected"/>
```
jupi3200: Does it indicate a user can upload multiple zip files?
Why do we need to support this?

lilian-hunt: data-multiple-caption="{count} files selected" this displays to the user how many files they have selected but we are only allowing them to select 1 (because the "multiple" tag is not there) so this tag can be removed

jupi3200 (approved):

PR #8, opened 2021-01-19
Add StopTooFarFromTripShape validation rule

KathleenQ:
**4 space** (rather than 2 space) is kept for all validation rules, notices and tests files to keep it consistent with all existing (rules, notices and tests) files for the new validator.
The corresponding notice is in PR #9 and tests are in PR #10 .

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
@@ -72,9 +72,13 @@ public class StopTooFarFromTripShapeValidator extends FileValidator {
             final List<GtfsStopTime> stopTimeList = tripIdStopTimeListEntry.getValue();
             final GtfsTrip trip = tripTable.byTripId(tripId);
             if (trip == null || stopTimeList == null || stopTimeList.isEmpty() || trip.shapeId() == null) {
+                // This rule only applies when all necessary fields exist.
                 continue;
             }
             final List<GtfsShape> shapeList = shapeTable.byShapeId(trip.shapeId());
+            if (shapeList == null || shapeList.isEmpty()) {
+                continue;
+            }
 
             // Create a polyline from the GTFS shapes data.
             ShapeFactory.LineStringBuilder lineBuilder = getShapeFactory().lineString();
@@ -88,7 +92,7 @@ public class StopTooFarFromTripShapeValidator extends FileValidator {
             // Check if each stop is within the buffer polygon.
             for (GtfsStopTime stopTime : stopTimeList) {
                 GtfsStop stop = stopTable.byStopId(stopTime.stopId());
-                if (stop == null || shapeList == null || shapeList.isEmpty()) {
+                if (stop == null) {
                     continue;
                 }
                 if (!(stop.locationType() == GtfsLocationType.STOP) && !(stop.locationType() == GtfsLocationType.BOARDING_AREA)) {
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
@@ -50,13 +50,17 @@ import static org.locationtech.spatial4j.context.SpatialContext.GEO;
  */
 @GtfsValidator
 public class StopTooFarFromTripShapeValidator extends FileValidator {
-    @Inject GtfsStopTimeTableContainer stopTimeTable;
+    @Inject
+    GtfsStopTimeTableContainer stopTimeTable;
 
-    @Inject GtfsTripTableContainer tripTable;
+    @Inject
+    GtfsTripTableContainer tripTable;
 
-    @Inject GtfsShapeTableContainer shapeTable;
+    @Inject
+    GtfsShapeTableContainer shapeTable;
 
-    @Inject GtfsStopTableContainer stopTable;
+    @Inject
+    GtfsStopTableContainer stopTable;
 
     // Spatial operation buffer values
     public static final double TRIP_BUFFER_METERS = 100;
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
@@ -104,7 +104,7 @@ public class StopTooFarFromTripShapeValidator extends FileValidator {
                 // Check whether the stop position is within an acceptable distance threshold from the trip shape.
                 Point p = getShapeFactory().pointXY(stop.stopLon(), stop.stopLat());
                 if (!shapeBuffer.relate(p).equals(SpatialRelation.CONTAINS)) {
-                    noticeContainer.addNotice(new StopTooFarFromTripShapeNotice(stopTime, trip, TRIP_BUFFER_METERS));
+                    noticeContainer.addNotice(new StopTooFarFromTripShapeNotice(stop.stopId(), stopTime.stopSequence(), trip.tripId(), trip.shapeId(), TRIP_BUFFER_METERS));
                 }
             }
         }
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
@@ -100,8 +100,10 @@ public class StopTooFarFromTripShapeValidator extends FileValidator {
                     continue;
                 }
                 if (testedCache.contains(trip.shapeId() + stop.stopId())) {
+                    // Skip tested shape_id and stop_id pair.
                     continue;
                 }
+                // A given trip shape and a specific stop (for the trip) combines a unique String key that only need to be checked once.
                 testedCache.add(trip.shapeId() + stop.stopId());
                 // Check whether the stop position is within an acceptable distance threshold from the trip shape.
                 Point p = getShapeFactory().pointXY(stop.stopLon(), stop.stopLat());
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
@@ -16,12 +16,14 @@
 
 package org.mobilitydata.gtfsvalidator.validator;
 
+import static org.locationtech.spatial4j.context.SpatialContext.GEO;
+
 import com.google.common.collect.Multimaps;
-import java.util.List;
-import java.util.Set;
-import java.util.Map;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import org.locationtech.spatial4j.distance.DistanceUtils;
 import org.locationtech.spatial4j.shape.Point;
 import org.locationtech.spatial4j.shape.Shape;
@@ -30,6 +32,7 @@ import org.locationtech.spatial4j.shape.SpatialRelation;
 import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
 import org.mobilitydata.gtfsvalidator.annotation.Inject;
 import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.StopTooFarFromTripShapeNotice;
 import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
 import org.mobilitydata.gtfsvalidator.table.GtfsShape;
 import org.mobilitydata.gtfsvalidator.table.GtfsShapeTableContainer;
@@ -39,41 +42,33 @@ import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
 import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
 import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
 import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
-import org.mobilitydata.gtfsvalidator.notice.StopTooFarFromTripShapeNotice;
-
-import static org.locationtech.spatial4j.context.SpatialContext.GEO;
 
 /**
  * Validates that a stop is within a distance threshold for a trip shape.
- * <p>
- * Generated notices:
- * * StopTooFarFromTripShapeNotice
+ *
+ * <p>Generated notices: * StopTooFarFromTripShapeNotice
  */
 @GtfsValidator
 public class StopTooFarFromTripShapeValidator extends FileValidator {
-    @Inject
-    GtfsStopTimeTableContainer stopTimeTable;
+    @Inject GtfsStopTimeTableContainer stopTimeTable;
 
-    @Inject
-    GtfsTripTableContainer tripTable;
+    @Inject GtfsTripTableContainer tripTable;
 
-    @Inject
-    GtfsShapeTableContainer shapeTable;
+    @Inject GtfsShapeTableContainer shapeTable;
 
-    @Inject
-    GtfsStopTableContainer stopTable;
+    @Inject GtfsStopTableContainer stopTable;
 
     // Spatial operation buffer values
     static final double TRIP_BUFFER_METERS = 100;
-    private static final double TRIP_BUFFER_DEGREES = 
-        DistanceUtils.KM_TO_DEG * (TRIP_BUFFER_METERS / 1000.0d);
+    private static final double TRIP_BUFFER_DEGREES =
+            DistanceUtils.KM_TO_DEG * (TRIP_BUFFER_METERS / 1000.0d);
 
     @Override
     public void validate(NoticeContainer noticeContainer) {
         // Cache for previously tested shape_id and stop_id pairs - no need to test them more than once.
         final Map<String, Set<String>> testedCache = new HashMap<>();
         // Go through the pair of tripId and the corresponding stop time list one by one.
-        for (Map.Entry<String, List<GtfsStopTime>> tripIdStopTimeListEntry : 
+        for (Map.Entry<String, List<GtfsStopTime>> tripIdStopTimeListEntry :
                 Multimaps.asMap(stopTimeTable.byTripIdMap()).entrySet()) {
             final String tripId = tripIdStopTimeListEntry.getKey();
             final List<GtfsStopTime> stopTimeList = tripIdStopTimeListEntry.getValue();
@@ -94,7 +89,8 @@ public class StopTooFarFromTripShapeValidator extends FileValidator {
                 if (stop == null) {
                     continue;
                 }
-                if (stop.locationType() != GtfsLocationType.STOP && stop.locationType() != GtfsLocationType.BOARDING_AREA) {
+                if (stop.locationType() != GtfsLocationType.STOP
+                        && stop.locationType() != GtfsLocationType.BOARDING_AREA) {
                     // This rule only applies to stops of location_type "STOP" or "BOARDING_AREA".
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


KathleenQ:
@googlebot I signed it!

jupi3200 (commented):
Can you please add tests?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,116 @@
...
+    // Spatial operation buffer values
+    public static final double TRIP_BUFFER_METERS = 100;
+    private static final double TRIP_BUFFER_DEGREES = DistanceUtils.KM_TO_DEG * (TRIP_BUFFER_METERS / 1000.0d);
+
+    @Override
+    public void validate(NoticeContainer noticeContainer) {
+        // Cache for previously tested shape_id and stop_id pairs - no need to test them more than once.
+        final Set<String> testedCache = new HashSet<>();
+        // The maximum acceptance distance difference from the stop to the trip shape.
+        final double distanceThreshold = 1000.0;
```
lilian-hunt: Not sure, but should this be a class variable, like TRIP_BUFFER_METERS?

jupi3200: In meters/km/miles/feet/inch?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,116 @@
...
+        // Go through the pair of tripId and the corresponding stop time list one by one.
+        for (Map.Entry<String, List<GtfsStopTime>> tripIdStopTimeListEntry : Multimaps.asMap(stopTimeTable.byTripIdMap()).entrySet()) {
+            final String tripId = tripIdStopTimeListEntry.getKey();
+            final List<GtfsStopTime> stopTimeList = tripIdStopTimeListEntry.getValue();
+            final GtfsTrip trip = tripTable.byTripId(tripId);
+            if (trip == null || stopTimeList == null || stopTimeList.isEmpty() || trip.shapeId() == null) {
+                continue;
```
jupi3200, line 78: // Comment why

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,116 @@
...
+                lineBuilder.pointXY(shape.shapePtLon(), shape.shapePtLat());
+            }
+            Shape shapeLine = lineBuilder.build();
+            // Create the buffered version of the trip as a polygon.
+            Shape shapeBuffer = shapeLine.getBuffered(TRIP_BUFFER_DEGREES, shapeLine.getContext());
+
+            // Check if each stop is within the buffer polygon.
+            for (GtfsStopTime stopTime : stopTimeList) {
+                GtfsStop stop = stopTable.byStopId(stopTime.stopId());
+                if (stop == null || shapeList == null || shapeList.isEmpty()) {
```
jupi3200: shapeList does not change in the loop.
Do we need to check it so many times?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,116 @@
...
+                    continue;
+                }
+                if (!(stop.locationType() == GtfsLocationType.STOP) && !(stop.locationType() == GtfsLocationType.BOARDING_AREA)) {
+                    // This rule only applies to stops of location_type "STOP" or "BOARDING_AREA".
+                    continue;
+                }
+                if (testedCache.contains(trip.shapeId() + stop.stopId())) {
```
jupi3200: Why are we adding Ids?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,116 @@
...
+            final List<GtfsShape> shapeList = shapeTable.byShapeId(trip.shapeId());
+
+            // Create a polyline from the GTFS shapes data.
+            ShapeFactory.LineStringBuilder lineBuilder = getShapeFactory().lineString();
+            for (GtfsShape shape : shapeList) {
+                lineBuilder.pointXY(shape.shapePtLon(), shape.shapePtLat());
+            }
+            Shape shapeLine = lineBuilder.build();
+            // Create the buffered version of the trip as a polygon.
+            Shape shapeBuffer = shapeLine.getBuffered(TRIP_BUFFER_DEGREES, shapeLine.getContext());
```
jupi3200: Can this be extracted to a method called getExpectedStopsLocation?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: I'm sorry that I just realised this parameter has not been used in my final strategy. I have just deleted it.

KathleenQ: I'm sorry that I just realised this parameter has not been used in my final strategy. I have just deleted it.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: I added the comment "This rule only applies when all necessary fields exist." as explanation for the fields' check.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Thank you for figuring it out! Yes, we only need to check it once. I have changed the position of "shapeList == null || shapeList.isEmpty()" just after getting "shapeList".

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: All Ids are strings. We use "trip.shapeId() + stop.stopId()" to combine a unique String key which makes the trip shape and stop pair only need to be checked once.
I have added comments for "trip.shapeId() + stop.stopId()" related parts in codes.

jupi3200: What is the guarantee that trip.shapeId() + stop.stopId() is an appropriate unique identifier.

Is it possible to have
trip1.shapeId = 'abcd' ; stop1.stopId = 'ef'
trip2.shapeId = 'abc' ;  stop2.stopId = 'def'
?

If you used AutoValue for this, you could enjoy a helpful hash function.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Good idea! I have created a helper method for this part, while I named it as "createTripShapePolygonGivenThreshold".
This method creates an initial trip shape polygon given the Gtfs shapes data and then expands the shape according to the distance threshold between a stop and the initial trip shape polygon. So I do not consider the final polygon as expected stops locations, but as a trip shape polygon with maximum tolerance.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Thank you for the example! Yes, it could happen and the validator will miss some pairs.
I currently plan to use HashMap for the records rather than the Set of String - I think it is an easy way that can record pairs correctly and is also easy to check the duplicated pairs later.
BTW, I want to ask, if we use AutoValue here, does that mean we need to add the "AutoValue" annotations for the corresponding types (i.e. GtfsTrip, GtfsStop) in the original files, or we need to create new corresponding types?

KathleenQ: Done!

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,129 @@
...
+ */
+@GtfsValidator
+public class StopTooFarFromTripShapeValidator extends FileValidator {
+    @Inject GtfsStopTimeTableContainer stopTimeTable;
+
+    @Inject GtfsTripTableContainer tripTable;
+
+    @Inject GtfsShapeTableContainer shapeTable;
+
+    @Inject GtfsStopTableContainer stopTable;
```
tripd22, line 59: I think we should put the annotation on its own line for consistency with the existing validators e.g.
```
@Inject
GtfsStopTimeTableContainer stopTimeTable;
@Inject
GtfsTripTableContainer tripTable;
...
```

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,131 @@
...
+
+    // Spatial operation buffer values
+    public static final double TRIP_BUFFER_METERS = 100;
```
tripd22: Is this public for a reason?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public void validate(NoticeContainer noticeContainer) {
+        // Cache for previously tested shape_id and stop_id pairs - no need to test them more than once.
+        final Map<String, List<String>> testedCache = new HashMap<>();
+        // Go through the pair of tripId and the corresponding stop time list one by one.
+        for (Map.Entry<String, List<GtfsStopTime>> tripIdStopTimeListEntry : Multimaps.asMap(stopTimeTable.byTripIdMap()).entrySet()) {
+            final String tripId = tripIdStopTimeListEntry.getKey();
+            final List<GtfsStopTime> stopTimeList = tripIdStopTimeListEntry.getValue();
+            final GtfsTrip trip = tripTable.byTripId(tripId);
+            if (trip == null || stopTimeList == null || stopTimeList.isEmpty() || trip.shapeId() == null) {
```
tripd22: Can stopTimeList be null or empty here? I don't know, but it's possible that they can never be null or never be empty (or both) here because for a tripId to appear in the stopTimeTable, it must be associated with a stopTime.

If this is the case, I think it would be better to avoid checking for null / empty, because a reader might assume that it could be null / empty.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,131 @@
...
+                // This rule only applies when all necessary fields exist.
+                continue;
+            }
+            final List<GtfsShape> shapeList = shapeTable.byShapeId(trip.shapeId());
+            if (shapeList == null || shapeList.isEmpty()) {
```
tripd22: Can this be null _and_ empty?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+            // Get the trip shape polygon given the Gtfs shapes data and TRIP_BUFFER_DEGREES.
+            Shape tripShapeGivenThreshold = createTripShapePolygonGivenThreshold(shapeList);
+            // Check if each stop is within the buffer polygon.
+            for (GtfsStopTime stopTime : stopTimeList) {
+                GtfsStop stop = stopTable.byStopId(stopTime.stopId());
+                if (stop == null) {
+                    continue;
+                }
+                if (!(stop.locationType() == GtfsLocationType.STOP) && !(stop.locationType() == GtfsLocationType.BOARDING_AREA)) {
```
tripd22: You could make this easier to read like:
```
if (stop.locationType() != GtfsLocationType.STOP && stop.locationType() != GtfsLocationType.BOARDING_AREA) {
```

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,131 @@
...
+    @Inject GtfsStopTableContainer stopTable;
+
+    // Spatial operation buffer values
+    public static final double TRIP_BUFFER_METERS = 100;
+    private static final double TRIP_BUFFER_DEGREES = DistanceUtils.KM_TO_DEG * (TRIP_BUFFER_METERS / 1000.0d);
+
+    @Override
+    public void validate(NoticeContainer noticeContainer) {
+        // Cache for previously tested shape_id and stop_id pairs - no need to test them more than once.
+        final Map<String, List<String>> testedCache = new HashMap<>();
```
tripd22: Have you considered any alternative data structures? What might the pros/cons be?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
```diff
@@ -0,0 +1,131 @@
...
+                    // Skip tested shape_id and stop_id pair.
+                    continue;
+                } else {
+                    // Record the new shape_id and stop_id pair when shape_id has appeared.
+                    testedCache.get(trip.shapeId()).add(stop.stopId());
+                }
+                // Check whether the stop position is within an acceptable distance threshold from the trip shape.
+                Point p = getShapeFactory().pointXY(stop.stopLon(), stop.stopLat());
+                if (!tripShapeGivenThreshold.relate(p).equals(SpatialRelation.CONTAINS)) {
+                    noticeContainer.addNotice(new StopTooFarFromTripShapeNotice(stop.stopId(), stopTime.stopSequence(), trip.tripId(), trip.shapeId(), TRIP_BUFFER_METERS));
```
tripd22: Nit: please fix formatting

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Yes, thank you for pointing it out! I have already fixed that.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Yes. As the threshold, TRIP_BUFFER_METERS is considered important to let users know in the notice. So in the tests file, this value has been imported.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Thank you for the advice! I have fixed that.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: If the relationship among tables are complete and no empty table, then yes, there are some null or empty conditions that do not need to be checked for this rule. Can we assume that (and leave these conditions checked in other specific rules)?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Yes, I previously used Set<String> to record the shape_id and stop_id pair by connecting them together as a long string key. But problem could happen. For example, when shape_id is "ABC" and stop_id is "DEF", this pair is considered the same as shape_id is "AB" and stop_id is "CDEF", but they are actually different. Because shape_id and stop_id both do not have length limits or requirements, so no matter what we add inside their connecting combination, for example, add "STOP" before stop_id, I don't think it can logically handle this problem. 
Then I also considered to continue using the Set, but maybe transfer shape_id and stop_id into unique hash code and then "connect" the numbers together.
Compared to using Map<String, List<String>> (as what I do here), the advantage of using hash code Set is the storage structure is simpler, but it asks to transfer the ids into hash code first and then compare every time. For the Map, the disadvantage is it is "2D" rather than "1D" as a Set - which means we need to compare key and then the corresponding value each time, but we could compare whether the pair exists and add them at the same time, and there is no need to do further complex calculation. That's why I finally choose to use the Map rather than the Set.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: I have adjusted the format referring to the clang format result, while still keeping it 4 space be to consistent with other rules' files.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Yes. I now understand what you mean. I'm going to change List<String> to Set<String> to have smaller time complexity when checking the uniqueness of stop_id inside.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
jupi3200: https://stackoverflow.com/questions/33267820/how-to-use-visiblefortesting-for-pure-junit-tests

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Done!

KathleenQ:
I will merge #10 into #8 as the corresponding tests after all feedback in PR #10 has been resolved.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Yes, the null and empty checks for stopTimeList are unnecessary. I have deleted them.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: I deleted the check for "shapeList.isEmpty()". Because it seems like the table will return null when the corresponding key does not exist, no matter whether the returned stuff is a single value or a list. (The same idea as .get(key) method will return null if a map contains no mapping for the key.)

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: Thank you for the advice! I currently changed it to be **with no access modifier**, which means TRIP_BUFFER_METERS is only accessible within classes in the same package - one level above "private", while more secure than "protected" and "public". I hope that works.

jupi3200 (approved):

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopTooFarFromTripShapeValidator.java
KathleenQ: I have run google-java-format to fix formatting again.

KathleenQ:
@googlebot I fixed it.

PR #9, opened 2021-01-19
Add notice for StopTooFarFromTripShape validation rule

KathleenQ:
**4 space** (rather than 2 space) is kept for all rules, notices and tests files to keep the coding style consistent with all existing (rules, notices and tests) files for the new validator.
The corresponding validation rule is in PR #8.

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
@@ -16,9 +16,6 @@
 
 package org.mobilitydata.gtfsvalidator.notice;
 
-import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
-import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
-
 import com.google.common.collect.ImmutableMap;
 
 public class StopTooFarFromTripShapeNotice extends Notice {
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+
+import com.google.common.collect.ImmutableMap;
+
+public class StopTooFarFromTripShapeNotice extends Notice {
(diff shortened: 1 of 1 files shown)


gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
```diff
@@ -0,0 +1,39 @@
...
+import com.google.common.collect.ImmutableMap;
+
+public class StopTooFarFromTripShapeNotice extends Notice {
+    public StopTooFarFromTripShapeNotice(String stopId, int stopSequence, String tripId, String shapeId, double tripBufferMeters) {
+        super(ImmutableMap.of(
+            "stopId", stopId,
+            "stopSequence", stopSequence,
+            "tripId", tripId,
+            "shapeId", shapeId,
+            "tripBufferMeters", tripBufferMeters
```
jupi3200: If we wish to support visualizations we would need to add more fields. 
We can add it later, so please feel free to add a TODO.

KathleenQ: Thank you for the advice! Yes, I think that is what we will do later. I have already added the TODO for the file.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
```
tripd22: Do we need these imports here?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.common.collect.ImmutableMap;
+
+public class StopTooFarFromTripShapeNotice extends Notice {
+    // TODO: More fields (e.g. stop location & information related to trip shape polygon) will be added to support notice visualisation.
+    public StopTooFarFromTripShapeNotice(String stopId, int stopSequence, String tripId, String shapeId, double tripBufferMeters) {
```
tripd22: Have you run clang format?

tripd22 (commented):
Be careful with the ordering of your PRs - we always want our repo in a working state. If you submit PR#8 before PR#9, I think the StopTooFarFromTripShapeValidator won't compile because it depends on the StopTooFarFromTripShapeNotice.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
KathleenQ: I have adjusted the format referring to the clang format result, while still keeping the format consistent with other notice files.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
KathleenQ: Thank you for pointing it out! Yes, we don't need them. I have deleted them.

KathleenQ:
For the order of PRs, yes, I realised it is a problem. I should put this notice file PR#9 before the related validation rule PR#8 and tests PR#10, to make the validator work all the time. I will pay attention to that next time.

jupi3200 (approved):

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTooFarFromTripShapeNotice.java
KathleenQ: I have run google-java-format to fix formatting again.

PR #12, opened 2021-01-25
Auto-formatting the whole repository

KathleenQ:
I ran auto formatting for all files (that are already in the master branch) in the repository using [google-java-format](https://github.com/google/google-java-format). 2 spaces are kept for all files. 
It makes the format style consistent throughout our repo. (It is also what the new validator repo recently did, after we cloned it into our own repository.)
_p.s. I used IntelliJ plugin for the reformatting as recommended on google-java-format website._


KathleenQ:
I will have another PR with only reformatting on java files and paying attention to broken formats.

PR #14, opened 2021-01-27
Decreasing stop time distance notice

lilian-hunt:


aannie-y (commented):
lgtm

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/DecreasingStopTimeDistanceNotice.java
```diff
@@ -0,0 +1,46 @@
...
+  public DecreasingStopTimeDistanceNotice(
+      String tripId,
+      long csvRowNumber,
+      int stopSequence,
+      double shapeDistTraveled,
+      long prevCsvRowNumber,
+      int prevStopSequence,
+      double prevShapeDistTraveled) {
+    super(
+        new ImmutableMap.Builder<String, Object>()
```
KathleenQ, line 31: You can directly use "ImmutableMap.of(...)" here, to be consistent with all other notices.
No need to have ".put(...)" and ".build()" then.

lilian-hunt: No, not all notices are like this.  "ImmutableMap.of(...)" is limited to 5 entires. See DecreasingShapeDistanceNotice.java that uses the .put and .build

KathleenQ: Sorry that I did not realised that earlier.

jupi3200 (approved):

PR #16, opened 2021-01-27
Add TripNumberOfStops Rule

KathleenQ:
Here is the rule validates that each trip must have more than one stop to be meaningful.
I combined the rule, its corresponding tests and the very short notice files into the same PR.
All files have already been reformatted using google-java-format.

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/MeaninglessTripNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/MeaninglessTripNotice.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+public class MeaninglessTripNotice extends Notice {
+  public MeaninglessTripNotice(String tripId, long csvRowNumber) {
+    super(
+        ImmutableMap.of(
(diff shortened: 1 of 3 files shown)


aannie-y (commented):
Maybe add another test where you have 0 stops as well? But other than that, lgtm

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/MeaninglessTripNotice.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
```
lilian-hunt, line 18: Create a seperate PR for this file (this should be first, then create the validator file with tests as the 2nd PR)

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripNumberOfStopsValidator.java
```diff
@@ -0,0 +1,49 @@
...
+@GtfsValidator
+public class TripNumberOfStopsValidator extends FileValidator {
+  @Inject GtfsTripTableContainer tripTable;
+
+  @Inject GtfsStopTimeTableContainer stopTimeTable;
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    for (GtfsTrip trip : tripTable.getEntities()) {
+      String tripId = trip.tripId();
```
lilian-hunt, line 41: this could be final

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/MeaninglessTripNotice.java
KathleenQ: Sure. I will close this one and have another two PRs.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripNumberOfStopsValidator.java
KathleenQ: Thank you for pointing it out!

KathleenQ:
@aannie-y Will consider to add that. Thank you!

KathleenQ:
Will create two PRs for notice and rule with tests separately.

PR #18, opened 2021-01-27
Add rule and tests for TripNumberOfStopsValidator

KathleenQ:


gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripNumberOfStopsValidatorTest.java
```diff
@@ -0,0 +1,166 @@
...
+    // tripX has stopA, stopB, stopD in order
+    // tripY has stopB, stopC in order
+    // tripZ has only stopB - Meaningless Trip
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    TripNumberOfStopsValidator validator = new TripNumberOfStopsValidator();
+
+    // Create tripTable:
+    List<GtfsTrip> trips = new ArrayList<>();
+    trips.add(tripX);
+    trips.add(tripY);
```
jupi3200, line 138: How is tripY or tripX relevant to the test?

KathleenQ: These two are not directly relevant to the notice result, but I want to set the situation that we have several trips in record, while one of them is meaningless, and better not the one in the first line of the csv file. Thus we can check whether the validator can point out the meaningless trip among all trips and return the correct csvRowNumber along with the trip_id, which is more similar to real-life condition.

jupi3200: In this case it would be helpful if you had.a the following tests:

- Single valid trip
- Trip without stops
- Trip with a single stop
- Combination of all the previous three trips.

KathleenQ: When the stopTime variables are defined outside the test methods, their corresponding csvRowNumbers are set. I don't think it is correct if we have only one trip for a test, but the csvRowNumber is not 1. 
In this case, if we have the single trip first and then combination later, we need to set trips that with the same values of all other fields except their csvRowNumbers. It means the tests will have no common trips at all due to the csvRowNumber. 
Is that fine that we define all stopTimes inside the tests, rather than outside as variables (as what I do now)? Duplicates will happen in this case.

KathleenQ: Have just updated the new tests.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripNumberOfStopsValidatorTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+    List<GtfsTrip> trips = new ArrayList<>();
+    trips.add(tripY);
+    validator.tripTable = GtfsTripTableContainer.forEntities(trips, noticeContainer);
+
+    validator.stopTimeTable =
+        GtfsStopTimeTableContainer.forEntities(new ArrayList<>(), noticeContainer);
+
+    validator.validate(noticeContainer);
+    assertThat(noticeContainer.getNotices()).containsExactly(new MeaninglessTripNotice("tripY", 2));
```
jupi3200, line 107: What is the meaning of **2** can you possibly add /* paramname= */

jupi3200, line 107: What is the meaning of **2** can you possibly add /* paramname= */

jupi3200 (approved):

jupi3200 (approved):

PR #20, opened 2021-01-28
Missing trip edge notice

lilian-hunt:


gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/MissingTripEdgeStopTimeNotice.java
```diff
@@ -0,0 +1,40 @@
...
+// Trip edge = first or last stop 
+public class MissingTripEdgeStopTimeNotice extends Notice {
+  public MissingTripEdgeStopTimeNotice(
+      String fieldName,
+      String tripId, 
+      int csvRowNumber,
+      int stopSequence) {
+    super(
+        ImmutableMap.of(
+            "fieldName", fieldName,
```
KathleenQ: Maybe name it directly as "arrival_or_departure_time" to be clearer?

KathleenQ (commented):
Except for the field name advice, LGTM.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/MissingTripEdgeStopTimeNotice.java
lilian-hunt: Yep good idea, done now

jupi3200 (approved):

PR #21, opened 2021-01-28
Missing trip edge rule

lilian-hunt:


KathleenQ (commented):
LGTM

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidator.java
```diff
@@ -0,0 +1,69 @@
...
+import org.mobilitydata.gtfsvalidator.annotation.Inject;
+import org.mobilitydata.gtfsvalidator.notice.MissingTripEdgeStopTimeNotice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+
+/**
+ * Validates that trip edges (first and last stops) for a trip define both arrival and deperature
```
tripd22: Nit: 'departure'

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
```diff
@@ -0,0 +1,262 @@
...
+    List<GtfsTrip> trips = new ArrayList<>();
+    trips.add(trip1);
+    validator.tripTable = GtfsTripTableContainer.forEntities(trips, noticeContainer);
+
+    validator.validate(noticeContainer);
+    assertThat(noticeContainer.getNotices()).isEmpty();
+  }
+
+  @Test
+  public void missingDepartureTimesForStartAndEndStop() {
```
tripd22: I would lean towards having individual tests for each version of the notice, e.g. `missingDepartureTimeForStartStop`, `missingDepartureTimeForEndStop`.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidator.java
```diff
@@ -0,0 +1,69 @@
...
+  @Inject GtfsTripTableContainer tripTable;
+
+  @Inject GtfsStopTimeTableContainer stopTimeTable;
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    for (GtfsTrip trip : tripTable.getEntities()) {
+      final String tripId = trip.tripId();
+      List<GtfsStopTime> stopTimes = stopTimeTable.byTripId(tripId);
+      // Only check if this is a meaningly trip with at least start and end stop
```
tripd22: Nit: rephrase?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidator.java
```diff
@@ -0,0 +1,69 @@
...
+      if (stopTimes.size() >= 2) {
+        GtfsStopTime tripStartStop = stopTimes.get(0);
+        GtfsStopTime tripEndStop = stopTimes.get(stopTimes.size() - 1);
+        if (tripStartStop.arrivalTime() == null) {
+          noticeContainer.addNotice(new MissingTripEdgeStopTimeNotice(
+              "arrival_time", tripId, tripStartStop.csvRowNumber(), tripStartStop.stopSequence()));
+        }
+        if (tripStartStop.departureTime() == null) {
```
tripd22: I think Mobility Data might have [implemented this rule earlier today](https://github.com/MobilityData/gtfs-validator/pull/699/files)...

Not to worry, I think we should still implement it. But maybe take a quick look to see if there is anything you could borrow from their approach. E.g. instead of checking for null here, maybe use `!tripStartStop.hasArrivalTime()` (which seems to be the approach some of the other files take as well).

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
```diff
@@ -0,0 +1,262 @@
...
+import org.mobilitydata.gtfsvalidator.type.GtfsTime;
+
+@RunWith(JUnit4.class)
+public class TripEdgeArrivalDepartureTimeValidatorTest {
+  private final GtfsTrip trip1 =
+      new GtfsTrip.Builder().setCsvRowNumber(1).setTripId("trip1").build();
+  private final GtfsTrip trip2 =
+      new GtfsTrip.Builder().setCsvRowNumber(2).setTripId("trip2").build();
+
+  private final GtfsStopTime startStopTimeHasBoth =
```
tripd22: We could cut down on the length of this code using helper functions to create these StopTimes, which I think may make these tests easier to read (similar to the [Mobility Data approach](https://github.com/MobilityData/gtfs-validator/blob/d1a5587bb78e47baf981eeb9a30c938aa8e2dc40/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/MissingTripEdgeValidatorTest.java)). What are your thoughts here @jupi3200 ?

jupi3200: I think when the number of parameters is > 2 builders are much better (especially when the call site does not make identifying the param type). See go/java-practices/builders#when and http://go/ej3e-2

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidator.java
lilian-hunt: Sorry, changed now

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
lilian-hunt: Ok so I will keep the builder

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidator.java
lilian-hunt: Changed to check if the stop just has at least 1 stop, like Global Mobility's, and removed comment

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidator.java
lilian-hunt: Oh ok yep will double check before picking my next rule. Yep had a look and `if (stopTimesForTrip.isEmpty()) {
        return;
      }` and `!tripStartStop.hasArrivalTime()`

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
lilian-hunt: Ok, updated now with 8 unit tests

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
```diff
@@ -0,0 +1,333 @@
...
+  public void missingDepartureTimeForFirstStop() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    TripEdgeArrivalDepartureTimeValidator validator = new TripEdgeArrivalDepartureTimeValidator();
+
+    // Create stopTimeTable:
+    List<GtfsStopTime> stopTimes = new ArrayList<>();
+    stopTimes.add(startStopTimeHasArrivalTrip1);
+    stopTimes.add(middleStopTrip1);
+    stopTimes.add(endStopTimeHasBothTrip1);
+    validator.stopTimeTable = GtfsStopTimeTableContainer.forEntities(stopTimes, noticeContainer);
```
jupi3200: Could this code be simplified and more readable if you used something like:
validator.stopTimeTable = GtfsStopTimeTableContainer.forEntities(List.of(startStopTimeHasArrivalTrip1, middleStopTrip1, endStopTimeHasBothTrip1), noticeContainer);

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
```diff
@@ -0,0 +1,333 @@
...
+  private final GtfsStopTime startStopTimeHasBothTrip1 =
+      new GtfsStopTime.Builder()
+          .setCsvRowNumber(18)
+          .setTripId("trip1")
+          .setStopId("1001")
+          .setStopSequence(1)
+          .setArrivalTime(GtfsTime.fromString("12:02:34"))
+          .setDepartureTime(GtfsTime.fromString("12:03:34"))
+          .build();
+  private final GtfsStopTime startStopTimeHasArrivalTrip1 =
```
jupi3200: perhaps it would be more useful for understanding this test if we named such stop as startStopNoDepartureTrip1 ?

lilian-hunt: ok has been updated to reflect this for all variables

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
lilian-hunt: ok using Arrays.asList as List.of is only available with Java9

jupi3200 (approved):

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TripEdgeArrivalDepartureTimeValidatorTest.java
jupi3200: Thanks

PR #22, opened 2021-01-29
add task in build.gradle to run jarfile

aannie-y:
Does not work, gives error saying '--outputPathFormat build/template.soy.js" is not a valid option then shows the Jar file usage and all the parameters it accepts. That quoted parameter works when run manually


aannie-y:
Works now. Use `gradle runExecutableJar` to compile SoyToJsSrcCompiler.

jupi3200:
Can we close this PR?

aannie-y:
PR closed because we are now using Liferay's `buildSoy` plugin, which has been implemented already in another PR.

PR #23, opened 2021-01-29
Duplicate trip validator

lilian-hunt:
(Might not be useful to include into the validator because the duplicate_key notice already picks this up). 

I chose to use a map because I had to store the tripId and the corresponding csv row number. 

I then chose to use a HashMap because it is implemented as a hash table, and there is no ordering on keys or values.  This means that there is expected constant-time performance O(1) for operations like add(), and contains(). 

This is better than a TreeMap which has performance O(log(n)). And order is not important here so the red-black tree structure, (ordered by the key), which is not required. 

This means the space complexity is O(n) in relation to the number of trips, and time complexity O(n) because there is one iteration through the trip table.

### DIFF
--- a/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/DuplicateTripValidatorTest.java
+++ b/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/DuplicateTripValidatorTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2020 Google LLC, MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/DuplicateTripValidatorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.mobilitydata.gtfsvalidator.notice.DuplicateTripNotice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+
+@RunWith(JUnit4.class)
+public class DuplicateTripValidatorTest {
+  private final GtfsTrip trip1 =
+      new GtfsTrip.Builder().setCsvRowNumber(21).setTripId("trip1").build();
+  private final GtfsTrip trip2 =
```
jupi3200: perhaps it would be more readable if you names the variable 'trip1_line39'

lilian-hunt: updated to trip1_line21 and trip1_line39

jupi3200 (approved):

PR #24, opened 2021-01-29
Add "duplicate route long name and short name combination" rule

KathleenQ:
This PR adds a validator rule with its corresponding notice and tests for no route with duplicate route long name and short name combination.
Only "DuplicateRouteLongNameShortNameCombination" check is finished in the "RouteUniqueNamesValidator". "DuplicateRouteShortName" and "DuplicateRouteLongName" checks for the same agency will be added later - space is left using comments.

In "RouteUniqueNamesValidator" file, I choose to use HashMap to record checked routes - the combination name (i.e. routeLongName + routeShortName) is used as key, the corresponding value is the route. I use HashMap to find the corresponding checked route which has the same combination name as the current checking route with smaller time complexity O(1). We need to store the route as value because we want to output the detailed information of the route if the notice is produced. Storing the route directly, rather than storing all necessary information around the route in the Map can save space.

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.validator;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.annotation.Inject;
+import org.mobilitydata.gtfsvalidator.notice.DuplicateRouteLongNameShortNameCombinationNotice;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsRoute;
+import org.mobilitydata.gtfsvalidator.table.GtfsRouteTableContainer;
+
+/**
+ * Validates short or long names for a route are unique.
+ *
+ * <p>Generated notices:
+ *
+ * <ul>
+ *   <li>{@link DuplicateRouteLongNameShortNameCombinationNotice}
+ *   <li>{@link DuplicateRouteShortNameNotice}
+ *   <li>{@link DuplicateRouteLongNameNotice}
+ * </ul>
+ */
+@GtfsValidator
+public class RouteUniqueNamesValidator extends FileValidator {
+  @Inject GtfsRouteTableContainer routeTable;
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    final Map<String, GtfsRoute> routesByLongNameShortName = new HashMap<>();
+    for (GtfsRoute route : routeTable.getEntities()) {
+      String routeLongName = Optional.ofNullable(route.routeLongName()).orElse("");
+      String routeShortName = Optional.ofNullable(route.routeShortName()).orElse("");
+      if (routeLongName != "") {
+        // For DuplicateRouteLongNameNotice
+      }
+      if (routeShortName != "") {
+        // For DuplicateRouteShortNameNotice
+      }
+      // Checks uniqueness of combination of fields `route_long_name` and `route_short_name`
+      if (routesByLongNameShortName.containsKey(routeLongName + routeShortName)) {
+        GtfsRoute comparedRoute = routesByLongNameShortName.get(routeLongName + routeShortName);
+        noticeContainer.addNotice(
+            new DuplicateRouteLongNameShortNameCombinationNotice(
+                comparedRoute.routeId(),
+                Optional.ofNullable(comparedRoute.routeLongName()).orElse(""),
+                Optional.ofNullable(comparedRoute.routeShortName()).orElse(""),
+                route.routeId(),
+                routeLongName,
+                routeShortName));
+      } else {
+        routesByLongNameShortName.put(routeLongName + routeShortName, route);
+      }
+    }
+  }
+}
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
@@ -32,9 +32,9 @@ import org.mobilitydata.gtfsvalidator.table.GtfsRouteTableContainer;
  * <p>Generated notices:
  *
  * <ul>
- *   <li>{@link DuplicateRouteLongNameShortNameCombinationNotice}
- *   <li>{@link DuplicateRouteShortNameNotice}
  *   <li>{@link DuplicateRouteLongNameNotice}
+ *   <li>{@link DuplicateRouteShortNameNotice}
+ *   <li>{@link DuplicateRouteLongNameShortNameCombinationNotice}
  * </ul>
  */
 @GtfsValidator
@@ -48,10 +48,10 @@ public class RouteUniqueNamesValidator extends FileValidator {
       String routeLongName = Optional.ofNullable(route.routeLongName()).orElse("");
       String routeShortName = Optional.ofNullable(route.routeShortName()).orElse("");
       if (routeLongName != "") {
-        // For DuplicateRouteLongNameNotice
+        // TODO: For DuplicateRouteLongNameNotice
       }
       if (routeShortName != "") {
-        // For DuplicateRouteShortNameNotice
+        // TODO: For DuplicateRouteShortNameNotice
       }
       // Checks uniqueness of combination of fields `route_long_name` and `route_short_name`
       if (routesByLongNameShortName.containsKey(routeLongName + routeShortName)) {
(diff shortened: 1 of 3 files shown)


gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    final Map<String, GtfsRoute> routesByLongNameShortName = new HashMap<>();
+    for (GtfsRoute route : routeTable.getEntities()) {
+      String routeLongName = Optional.ofNullable(route.routeLongName()).orElse("");
+      String routeShortName = Optional.ofNullable(route.routeShortName()).orElse("");
+      if (routeLongName != "") {
+        // For DuplicateRouteLongNameNotice
+      }
```
jupi3200: This does nothing.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
```diff
@@ -0,0 +1,206 @@
...
+    validator.validate(noticeContainer);
+    assertThat(noticeContainer.getNotices())
+        .containsExactly(
+            new DuplicateRouteLongNameShortNameCombinationNotice(
+                "routeLongName_ab_ShortName_cd",
+                "ab",
+                "cd",
+                "routeLongName_abc_ShortName_d",
+                "abc",
+                "d"));
```
jupi3200: Is this a desired behaviour?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+  @Test
+  public void routeOnlyLongNameDifferentFromAnotherRouteOnlyLongName() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    RouteUniqueNamesValidator validator = new RouteUniqueNamesValidator();
+
+    List<GtfsRoute> routes = new ArrayList<>();
+    routes.add(routeOnlyLongName_abcd);
+    routes.add(routeOnlyLongName_alpha);
+    validator.routeTable = GtfsRouteTableContainer.forEntities(routes, noticeContainer);
```
jupi3200: Can you use Arrays.asList to make the code more readable?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
```diff
@@ -0,0 +1,206 @@
...
+          .setCsvRowNumber(6)
+          .setRouteId("routeLongName_abc_ShortName_d")
+          .setRouteLongName("abc")
+          .setRouteShortName("d")
+          .build();
+
+  @Test
+  public void routeOnlyLongNameEqualsAnotherRouteOnlyShortName() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    RouteUniqueNamesValidator validator = new RouteUniqueNamesValidator();
```
jupi3200: These can be common fields to avoid the code duplication.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
```diff
@@ -0,0 +1,206 @@
...
+    List<GtfsRoute> routes = new ArrayList<>();
+    routes.add(routeOnlyLongName_abcd);
+    routes.add(routeOnlyShortName_abcd);
+    validator.routeTable = GtfsRouteTableContainer.forEntities(routes, noticeContainer);
+
+    validator.validate(noticeContainer);
+    assertThat(noticeContainer.getNotices())
+        .containsExactly(
+            new DuplicateRouteLongNameShortNameCombinationNotice(
+                "routeOnlyLongName_abcd", "abcd", "", "routeOnlyShortName_abcd", "", "abcd"));
```
jupi3200: please add /* paramName= */ comments to make the test more readable.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
KathleenQ: Yes. The two-condition checking is prepared for two other rules in this validator. I will change it as TODO.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
KathleenQ: Yes. Because a route with LongName "ab" and ShortName "cd" has the name combination "abcd", and it is the same with a route with LongName "abc" and ShortName "d", which should generate a warning for DuplicateRouteLongNameShortNameCombination.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
KathleenQ: Sure. Good idea! Has been fixed!

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
KathleenQ: Good idea! Done.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
jupi3200: I am not sure why this requirement needs to be true.
Looking at https://developers.google.com/transit/gtfs/reference#routestxt There does not seem to be such explicit requirement.
If the old validator implemented such a rule we can enforce it for compatibility.

@tripd22 Can you please check if we want to enforce such requirement?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
KathleenQ: Good idea! I have just fixed that.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
KathleenQ: Yes, the rule logic comes from "UniqueRouteLongNameRouteShortNameCombination" Warning in the old validator. 
I thought it is meaningful because we only check the uniqueness of short name and long name of route respectively for the same agency _(in the left two rules that I have not implemented in the validator)_, while never checking what will happen if the agency is not the same. 
This rule makes sure the combination of information from route's name is unique and meaningful. Also, when the route only has long name or short name (which will happen from time to time), the only name is unique among all names (both long and short ones).

tripd22: I think what we actually want to do is ensure that the combination of short name + long name + route type + agency id is unique.

This is what our C++ validator currently checks (though it used to check just the short name + long name combination). However, it doesn't concatenate the strings to do this. Instead, it basically hashes the 4 individual pieces of data together, and ensures the hash is unique.

So {"ab", "cd", 1, "agency_a"} and {"abc", "d", 1, "agency_a"} would be considered different.

You can probably use a custom class that holds these pieces of data to replicate this, and hash that (unless you know a better approach @jupi3200? I'll send you the C++ code snippet where we avoid creating a data structure.).


Note: The old mobility data validator also tests that short name is unique _and_ long name is unique. Our C++ validator doesn't test for this, and it isn't part of the spec, so I don't think we need to do this.

KathleenQ: Sure. I will then wait to adjust the validator logic :) rather than further implementing the other two rules.

tripd22: @jupi3200 suggested [autovalue](https://www.baeldung.com/introduction-to-autovalue), which looks like a better approach than mine!

tripd22: [`Objects.hash(long_name, short_name, route_type, agency_id)`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-) might be useful as well - that way you don't need to create a new object to get a hash of the values.

KathleenQ: Thank you! I'm going to change to that.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,72 @@
...
+      String routeLongName = Optional.ofNullable(route.routeLongName()).orElse("");
+      String routeShortName = Optional.ofNullable(route.routeShortName()).orElse("");
+      if (routeLongName != "") {
+        // TODO: For DuplicateRouteLongNameNotice
+      }
+      if (routeShortName != "") {
+        // TODO: For DuplicateRouteShortNameNotice
+      }
+      // Checks uniqueness of combination of fields `route_long_name` and `route_short_name`
+      if (routesByLongNameShortName.containsKey(routeLongName + routeShortName)) {
```
jupi3200, line 57: When will this break? Can you avoid it?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidatorTest.java
KathleenQ: I used this idea and changed the validator. Because almost all files were changed, I created another new PR for the new version - PR #28 .

KathleenQ:
The validator idea has been changed. Now, we check whether the combination of names, type and agency ID for a route is unique, rather than the String combination of only long name and short name of a route, referring from rules in the C++ validator. The change is meaningful.
Because all files have been changed entirely, I close this PR and new changes are in PR #28 .

PR #25, opened 2021-02-02
create notice, validator, tests for route has no duplicate stops

lilian-hunt:
Checks all trips and ensures that consecutive stops do not have the same name. If they do a notice is generated. To avoid repeated notices, I have cached the routeId, stopId1 and stopId2, and only generate a notice for each route once. By using a HashSet this means that the lookup time is O(1) and the overall time complexity of the validator is O(n), as it iterates through each trip's stops once. The space complexity is also O(n) for the hash set. 

This is better than a TreeMap which has performance O(log(n)). And order is not important here so the red-black tree structure, (ordered by the key), which is not required.


gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidatorTest.java
```diff
@@ -0,0 +1,191 @@
...
+    // Create stopTimeTable:
+    validator.stopTimeTable = GtfsStopTimeTableContainer.forEntities(
+        Arrays.asList(stopTimeTrip1_Stop1, stopTimeTrip1_Stop2, stopTimeTrip2_Stop1, stopTimeTrip2_Stop2), noticeContainer);
+
+    // Create stopTable:
+    validator.stopTable =
+        GtfsStopTableContainer.forEntities(Arrays.asList(stopId1001, stopId1002), noticeContainer);
+    
+    validator.validate(noticeContainer);
+    System.out.println(noticeContainer.getNotices());
```
aannie-y: Do you need to print out the notice container? Or was this line just left over from debugging?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidatorTest.java
```diff
@@ -0,0 +1,191 @@
...
+    validator.stopTimeTable = GtfsStopTimeTableContainer.forEntities(
+        Arrays.asList(stopTimeTrip1_Stop1, stopTimeTrip1_Stop2, stopTimeTrip2_Stop1, stopTimeTrip2_Stop2, stopTimeTrip3_Stop1, stopTimeTrip3_Stop2), noticeContainer);
+
+    // Create stopTable:
+    validator.stopTable =
+        GtfsStopTableContainer.forEntities(Arrays.asList(stopId1001, stopId1002), noticeContainer);
+    
+    validator.validate(noticeContainer);
+    System.out.println(noticeContainer.getNotices().size());
+    System.out.println(noticeContainer.getNotices());
```
aannie-y: Same as above with the printing. Remove them if you don't need to print.

aannie-y (commented):
Other than the few sysout lines, lgtm!

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidatorTest.java
lilian-hunt: Sorry yes, should have cleaned this up before. Gone now.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidatorTest.java
lilian-hunt: Removed :)

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidator.java
```diff
@@ -0,0 +1,89 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+
+/**
+ * Validates that for each route, that consecutive stop times do not have the same stop name.
+ * It checks this for all trips.
+ * [Warning]
+ *
+ * <p>Generated notice: {@link DecreasingStopTimeDistanceNotice}.
```
tripd22: Nit: Update the type of notice.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteWithDuplicateStopNotice.java
```diff
@@ -0,0 +1,38 @@
...
+  public RouteWithDuplicateStopNotice(String stopName, String stopId1, long csvRowNumberStop1,
+      String stopId2, long csvRowNumberStop2, String routeId) {
+    super(new ImmutableMap.Builder<String, Object>()
+              .put("stopName", stopName)
+              .put("stopId1", stopId1)
+              .put("csvRowNumberStop1", csvRowNumberStop1)
+              .put("stopId2", stopId2)
+              .put("csvRowNumberStop2", csvRowNumberStop2)
+              .put("routeId", routeId)
+              .build());
```
tripd22: I think this notice should probably contain an `example_trip_id` in which the duplicate stop occurs, or contain the stop_times that triggered the notice.

If we don't include the trip id where we saw the error, the user might have to look through all their trips (+ stop times) that are associated with that route to find the error. (Note that 2 trips on the same route don't necessarily need to follow the same stops).

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidator.java
```diff
@@ -0,0 +1,89 @@
...
+import org.mobilitydata.gtfsvalidator.notice.RouteWithDuplicateStopNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTime;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTimeTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+
+/**
+ * Validates that for each route, that consecutive stop times do not have the same stop name.
```
tripd22: for each _trip_

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteWithDuplicateStopNotice.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+
+public class RouteWithDuplicateStopNotice extends Notice {
```
tripd22: I think `TripWithDuplicateStopNotice` may be a better name, as that is what is actually causing the error.

The reason we use route ID to deduplicate the notices is that if we have 1 route with 100 identical trips that all have the same error, we don't want to spam the user with warnings.

However, the error isn't actually in the route. It could be caused by a few things, including:
- 2 different stops in stops.txt having the same name
- The order of stops being incorrect for a single trip in stop_times.txt
- etc.

The trip + the offending stop times in the trip are probably the most important things for the user to be able to debug the issue.

lilian-hunt: But if this is the case then shouldn't we probably be generating a notice for each trip that so that they can debug the issues for all trips? Because otherwise how would they know that the other trips also have the same error.

tripd22: Yeah, the deduplication based on route ID is one of those 'more art than science' additions to the rule.

The idea is that if this notice appears for the same stops in many trips on the same route, then the stops are probably correct, but should be named differently. E.g. the feed has named two different stops in a city like `--->CBD--->CBD--->`, which would confuse users. If this is the case, then the user only needs to fix 'one' issue - the name of the stops in stops.txt.

These additions to rules are generally based on real data we see in GTFS feeds, + feedback from users of our validator. In particular, this addition was introduced because we were generating a _lot_ of erroneous warnings for this rule, and deduplicating based on route ID gave a better experience to users (although I agree it's definitely not perfect).

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidator.java
lilian-hunt: Yep updated now

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteWithDuplicateStopNotice.java
lilian-hunt: Good point. I have added an exampleTripId field to the notice now

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteHasNoDuplicateStopsValidator.java
lilian-hunt: Changed and explains it will only generate one notice per route.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteWithDuplicateStopNotice.java
lilian-hunt: Ok yes I see, this makes sense. I have changed the names of the files to reflect this now.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/TripHasNoDuplicateStopsValidator.java
```diff
@@ -0,0 +1,91 @@
...
+          } else {
+            combinedStopKey = Objects.hash(trip.routeId(), stop.stopId(), prevStopId);
+          }
+          if (prevStopName.equals(stop.stopName())) {
+            // Only create the notice once for each combination of stops
+            if (!seenDuplicateStops.contains(combinedStopKey)) {
+              noticeContainer.addNotice(
+                  new TripWithDuplicateStopNotice(stop.stopName(), prevStopId, prevCsvRowNumber,
+                      stop.stopId(), stop.csvRowNumber(), trip.routeId(), trip.tripId()));
+            }
```
jupi3200: How are you handling hash collision?

lilian-hunt: In Java8 HashMap collisions are handled automatically by a balanced tree instead of a linked list for storing collided entries. Which in the worst case results in performance of O(log n) for get().

jupi3200: Let's discuss this in our 1:1, I might learn something new.

lilian-hunt: Now using autovalue to avoid using Objects.hash that could possibly still result in hash collisions.

jupi3200 (approved):

PR #27, opened 2021-02-03
Javascript unit test base implementation

jupi3200:
Added initial support for using karma to execute unit test on Javascript code.

This tests will be automatically executed when running gradle :webapp:test

### DIFF
--- a/webapp/src/main/webapp/test/output.test.js
+++ b/webapp/src/main/webapp/test/output.test.js
@@ -1,0 +1,26 @@
+/*
+ * Unit tests for output.js
+ */
+describe('Output', function() {
+
+  // inject the HTML fixture for the tests
+  beforeEach(function() {
+    var fixture = '<div id="fixture"><div id="warning"></div></div>';
+
+    document.body.insertAdjacentHTML(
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/test/output.test.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+describe('Output', function() {
+
+  // inject the HTML fixture for the tests
+  beforeEach(function() {
+    var fixture = '<div id="fixture"><div id="warning"></div></div>';
+
+    document.body.insertAdjacentHTML(
+      'afterbegin', 
+      fixture);
```
aannie-y, line 12: Could you please explain what this does? What's the 'afterbegin' part?

aannie-y (commented):
lgtm

webapp/src/main/webapp/test/output.test.js
jupi3200: Following http://www.bradoncode.com/blog/2015/02/27/karma-tutorial/

tripd22 (approved):

PR #28, opened 2021-02-03
Add RouteUniqueNames Validator (with tests)

KathleenQ:
The rule validates whether the combination of names, type and agency ID for a route is unique.

A storage with type Map<Integer, List<Object>> is created to store the checked route information. Hash code is used as key of the Map. I choose to use HashMap, because it will be much easier to find the corresponding route information for the specific hash code - time complexity is only O(1). 
Values of the HashMap is a List with the long name, short name, integer value representation of route type and agency ID. I choose List rather than Set is because the order of these fields are really important for comparison. Why I choose List instead of Quintet (i.e. 5-element Tuple) is because with the same complexity for the operations that we will use, operations for list is easier (e.g. we can add all elements at one time using Arrays.asList()) and we can add extra elements if necessary later easier, rather than changing the tuple name again (i.e. from Quintet to another tuple data structure with more elements). 

In addition, why I store and check each field is exactly the same after two routes have the same hash code is to avoid when the number of route data is too large (which could happen in some extreme cases in real application), two routes with different information may have the same hash code. This condition is rare, but is a potential bug, so I fixed it.


aannie-y (commented):
lgtm

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteUniqueNamesNotice.java
```diff
@@ -0,0 +1,45 @@
...
+      String agencyId) {
+    super(
+        new ImmutableMap.Builder<String, Object>()
+            .put("routeId", routeId)
+            .put("comparedRouteId", comparedRouteId)
+            .put("routeLongName", routeLongName)
+            .put("routeShortName", routeShortName)
+            .put("routeType", routeType)
+            .put("agencyId", agencyId)
+            .build());
```
jupi3200: Should we add line number?

KathleenQ: Sure. Good idea! I will add that.

KathleenQ: Done.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,87 @@
...
+              new RouteUniqueNamesNotice(
+                  route.routeId(),
+                  route.csvRowNumber(),
+                  /* comparedRouteId= */ storedRouteInfo.get(0).toString(),
+                  /* comparedRouteCsvRowNumber= */ (long) (storedRouteInfo.get(5)),
+                  routeLongName,
+                  routeShortName,
+                  routeType,
+                  agencyId));
+        }
```
jupi3200: else?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,87 @@
...
+      } else {
+        List<Object> routeInfo =
+            new ArrayList<>(
+                Arrays.asList(
+                    route.routeId(),
+                    routeLongName,
+                    routeShortName,
+                    routeType.getNumber(),
+                    agencyId,
+                    route.csvRowNumber()));
```
jupi3200: The code could be much more simplified if you used something like:

```
@AutoValue
abstract class RouteIdentifier {
    static RouteIdentifier create(String shortName,String longName, GtfsRouteType routeType, String agencyId) {
        return new AutoValue_RouteIdentifier(shortName, longName,routeType,agencyId);
    }

    abstract String shortName();
    abstract String longName();
    abstract GtfsRouteType routeType();
    abstract String agencyId();
}
```

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
KathleenQ: Yes. Thank you for pointing it out! In the "else" case, the corresponding route also needs to be stored into the HashMap.
The key of HashMap is still the hash code, while the values can become the list of the list of object, though there is usually only one list of object as the value.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
KathleenQ: Thank you for the advice! I am considering that.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,87 @@
...
+@GtfsValidator
+public class RouteUniqueNamesValidator extends FileValidator {
+  @Inject GtfsRouteTableContainer routeTable;
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    final Map<Integer, List<Object>> testedRoutes = new HashMap<>();
+    for (GtfsRoute route : routeTable.getEntities()) {
+      // Grab necessary route information and transfer null to empty String to make later comparison
+      // and notice genration easier
```
tripd22: Nit: _generation_

KathleenQ: Thank you pointing out the typo! I've fixed it and will push with the latest version soon.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
KathleenQ: Done. I have used autovalue now.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
KathleenQ: Fixed!

KathleenQ: Now I have changed the autovalue as the key of HashMap, and the problem no longer exists :) Because we do not need to consider the bug related to hash code generation and large data size.

KathleenQ:
I have now changed to use AutoValue rather than hash code as the key to avoid the problem that when the route data size is too large, we may have two routes with the same hash codes.
The value of HashMap is still a List of Object. Now, it stores the route_id and route_csv_row_number, which will be printed out in the notice. Why I choose List here is because the elements' order matters (no Set here) and List can be extended more easier than Tuple (tuples with different lengths have different names) without explicit disadvantages for the current use.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,93 @@
...
+              Optional.ofNullable(route.agencyId()).orElse(""));
+
+      // If the combination of names, type and agency ID for a route appeared before, a notice is
+      // generated
+      if (testedRoutes.containsKey(routeIdentifier)) {
+        noticeContainer.addNotice(
+            new RouteUniqueNamesNotice(
+                route.routeId(),
+                route.csvRowNumber(),
+                /* comparedRouteId= */ testedRoutes.get(routeIdentifier).get(0).toString(),
```
jupi3200, line 80: is the toString() needed?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/RouteUniqueNamesValidator.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * <p>Generated notices: {@link RouteUniqueNamesNotice}.
+ */
+@GtfsValidator
+public class RouteUniqueNamesValidator extends FileValidator {
+  @Inject GtfsRouteTableContainer routeTable;
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    final Map<RouteIdentifier, List<Object>> testedRoutes = new HashMap<>();
```
jupi3200: Can we avoid the usage of List<Object>?
Using List<Object> reduces the ability of the compiler to check we have the expected set of details (especially when the used object is ArrayList.

this could be replaced by GtfsRoute of some AutoValue class containing routeId and csvRowNumber.

KathleenQ: Sure, I will do that.

KathleenQ: I have now changed it from the List of Object to another AutoValue to store routeId and csvRowNumber, as the value of the HashMap.

jupi3200 (approved):

PR #29, opened 2021-02-03
Stops Too Close Validator

lilian-hunt:
This rule checks that all stops in stops.txt are sufficiently distanced from each other. If they are too close it generates a warning. A naive solution to this problem would be to compare each stop to every other stop O(n^2). There are more complex algorithms that use a spatial index (a data structure that allows for accessing a spatial object efficiently, which average search O(log n) but worst case is O(n)). However, the added space complexity and time spent to create the spatial index is not necessary for this problem where there is a finite number of stops. 

This solution is a slight improvement on the naive solution. It sorts the stops by latitude first and then compares, which is O(n log n) using merge sort. If the stops are greater than 0.00005 degrees away, approximately 5m away based on the latitude, the loop breaks and we move onto the next stop. If all the stops are at the same location this has the worst case running time of O(n^2) but in practice it should improve the running time considerably. Resulting in a running time between O(n) and O(n^2).

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
@@ -40,18 +40,16 @@ import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
 @GtfsValidator
 public class StopsTooCloseValidator extends FileValidator {
   private static final int KILOMETER_TO_METER_CONVERSION_FACTOR = 1000;
-  /**
-   * At equator 1 degree = 110.567 kilometers.
-   * 0.00005 degrees = (110.567km * 1000) * 0.00005 = 5.52835 m
-   * At poles 1 degree = 111.699 kilometers.
-   * 0.00005 degrees = (111.699km * 1000) * 0.00005 = 5.58495 m
-   *
-   * So the latitude buffer is 0.00005 degrees to check stops within at least 5.52835 m
-   */
-  private static final double LAT_BUFFER = 0.00005;
 
   // If stops are closer than 5m based on distance calculated, generate a notice
   static final int DIST_BUFFER_METERS = 5;
+
+  /**
+   * Convert DIST_BUFFER_METERS to degrees of latitude. Converts a distance in the units of the
+   * radius to degrees (360 degrees are in a circle). Assumming spherical earth model
+   */
+  private static final double LAT_BUFFER = DistanceUtils.dist2Degrees(DIST_BUFFER_METERS,
+      DistanceUtils.EARTH_EQUATORIAL_RADIUS_KM* KILOMETER_TO_METER_CONVERSION_FACTOR);
 
   @Inject GtfsStopTableContainer stopTable;
 
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.validator;
+
+import static org.locationtech.spatial4j.context.SpatialContext.GEO;
+
+import com.google.common.collect.Multimaps;
+import java.lang.Double;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.locationtech.spatial4j.distance.DistanceCalculator;
+import org.locationtech.spatial4j.distance.DistanceUtils;
+import org.locationtech.spatial4j.shape.Point;
+import org.locationtech.spatial4j.shape.ShapeFactory;
+import org.mobilitydata.gtfsvalidator.annotation.GtfsValidator;
+import org.mobilitydata.gtfsvalidator.annotation.Inject;
+import org.mobilitydata.gtfsvalidator.notice.NoticeContainer;
+import org.mobilitydata.gtfsvalidator.notice.StopsTooCloseNotice;
+import org.mobilitydata.gtfsvalidator.table.GtfsLocationType;
+import org.mobilitydata.gtfsvalidator.table.GtfsStop;
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+
+/**
+ * Validates that two stops are not too close to each other.
+ *
+ * <p>Generated notice: {@link StopsTooCloseNotice}.
+ */
+@GtfsValidator
+public class StopsTooCloseValidator extends FileValidator {
+  static final int KILOMETER_TO_METER_CONVERSION_FACTOR = 1000;
+  /**
+  * At equator 1 degree = 110.567 kilometers.
+  * 0.00005 degrees = (110.567km * 1000) * 0.00005 = 5.52835 m
+  * At poles 1 degree = 111.699 kilometers.
+  * 0.00005 degrees = (111.699km * 1000) * 0.00005 = 5.58495 m
+  *
+  * So the latitude buffer is 0.00005 degrees to check stops within at least 5.52835 m 
+  */
+  private static final double LAT_BUFFER = 0.00005;
+
+  // If stops are closer than 5m based on distance calculated, generate a notice
+  static final int DIST_BUFFER_METERS = 5;
+
+  @Inject GtfsStopTableContainer stopTable;
+
+  private static ShapeFactory getShapeFactory() {
+    return GEO.getShapeFactory();
+  }
+
+  private static DistanceCalculator getDistanceCalculator() {
+    return GEO.getDistCalc();
+  }
+
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    List<GtfsStop> allStops = stopTable.getEntities();
+    Collections.sort(
+        allStops, (stop1, stop2) -> { return Double.compare(stop1.stopLat(), stop2.stopLat()); });
+
+    // Set up to calculate distance
+    final ShapeFactory shapeFactory = getShapeFactory();
+    final DistanceCalculator distanceCalculator = getDistanceCalculator();
+
+    for (int i = 0; i < allStops.size(); i++) {
+      GtfsStop stop1 = allStops.get(i);
+      Point pointStop1 = shapeFactory.pointXY(stop1.stopLon(), stop1.stopLat());
+
+      for (int j = i + 1; j < allStops.size(); j++) {
+        GtfsStop stop2 = allStops.get(j);
+        if (stop1.stopLat() - stop2.stopLat() < LAT_BUFFER) {
+          Point pointStop2 = shapeFactory.pointXY(stop2.stopLon(), stop2.stopLat());
+
+          int distBetweenStopsMeters =
+              (int) (DistanceUtils.DEG_TO_KM * distanceCalculator.distance(pointStop1, pointStop2)
+                  * KILOMETER_TO_METER_CONVERSION_FACTOR);
+
+          if (distBetweenStopsMeters < DIST_BUFFER_METERS) {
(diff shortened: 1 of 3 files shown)


gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
```diff
@@ -0,0 +1,103 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsStopTableContainer;
+
+/**
+ * Validates that two stops are not too close to each other.
+ *
+ * <p>Generated notice: {@link StopsTooCloseNotice}.
+ */
+@GtfsValidator
+public class StopsTooCloseValidator extends FileValidator {
+  static final int KILOMETER_TO_METER_CONVERSION_FACTOR = 1000;
```
KathleenQ: Does it have some special use that you choose to have no modifier rather than "private"?

KathleenQ (commented):
LGTM in general.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
lilian-hunt: I am not using outside of the class will add private.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
```diff
@@ -0,0 +1,103 @@
...
+  private static final int KILOMETER_TO_METER_CONVERSION_FACTOR = 1000;
+  /**
+  * At equator 1 degree = 110.567 kilometers.
+  * 0.00005 degrees = (110.567km * 1000) * 0.00005 = 5.52835 m
+  * At poles 1 degree = 111.699 kilometers.
+  * 0.00005 degrees = (111.699km * 1000) * 0.00005 = 5.58495 m
+  *
+  * So the latitude buffer is 0.00005 degrees to check stops within at least 5.52835 m 
+  */
+  private static final double LAT_BUFFER = 0.00005;
```
jupi3200: Could this value be derived from DIST_BUFFER_METERS (possibly using
https://locationtech.github.io/spatial4j/apidocs/org/locationtech/spatial4j/distance/DistanceUtils.html )?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
```diff
@@ -0,0 +1,103 @@
...
+    final ShapeFactory shapeFactory = getShapeFactory();
+    final DistanceCalculator distanceCalculator = getDistanceCalculator();
+
+    for (int i = 0; i < allStops.size(); i++) {
+      GtfsStop stop1 = allStops.get(i);
+      Point pointStop1 = shapeFactory.pointXY(stop1.stopLon(), stop1.stopLat());
+
+      for (int j = i + 1; j < allStops.size(); j++) {
+        GtfsStop stop2 = allStops.get(j);
+        if (stop1.stopLat() - stop2.stopLat() < LAT_BUFFER) {
```
jupi3200: Should this be always true?
If allStops is sorted by latitude in ascending order, and stop1 is before stop2, then stop1.stopLat() <= stop2.stopLat()

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
```diff
@@ -0,0 +1,103 @@
...
+          Point pointStop2 = shapeFactory.pointXY(stop2.stopLon(), stop2.stopLat());
+
+          int distBetweenStopsMeters =
+              (int) (DistanceUtils.DEG_TO_KM * distanceCalculator.distance(pointStop1, pointStop2)
```
jupi3200: Why are we calculating in int?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+  public void stopsSpacedOutShouldNotGenerateNotice() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    StopsTooCloseValidator validator = new StopsTooCloseValidator();
+
+    // Create stopTable:
+    validator.stopTable =
+        GtfsStopTableContainer.forEntities(Arrays.asList(stop1, stop2), noticeContainer);
+
+    validator.validate(noticeContainer);
+    System.out.println(noticeContainer.exportJson());
```
jupi3200: Why are we including a println?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+    assertThat(noticeContainer.getNotices()).isEmpty();
+  }
+
+  @Test
+  public void stopsTooCloseGenerateNotice() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    StopsTooCloseValidator validator = new StopsTooCloseValidator();
```
jupi3200: We can set it as a common code instead of repeating.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidatorTest.java
lilian-hunt: Sorry I should have checked and removed it before.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidatorTest.java
lilian-hunt: As in put these 2 lines in like a @Before test?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
lilian-hunt: Good catch. I have reversed them now

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
lilian-hunt: This is how I saw it done in previous mobility data code, but casting to int reduces precision so I have removed it now and left it as a double.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
lilian-hunt: They only have DistanceUtils.calcLonDegreesAtLat(), but not the opposite to calculate Lat Degrees. However, this takes the distance in degrees and not meters. Then there is dist2Degrees(double dist, double radius) but requires the radius of the earth which varies as well. So it doesn't appear there is a straight forward calculation.

jupi3200: I think you could use a spherical earth model for this calculation, (using https://locationtech.github.io/spatial4j/apidocs/org/locationtech/spatial4j/distance/DistanceUtils.html#EARTH_EQUATORIAL_RADIUS_KM )

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidatorTest.java
jupi3200: I would have converted there two variable to be an instance variable.

lilian-hunt: I would still have to clear the notice container in each test. But can make the validator a instance variable

lilian-hunt: Ok updated

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
lilian-hunt: I think it makes more sense to have the buffer as latitude instead of longitude: 

Degrees of latitude are parallel so the distance between each degree remains almost constant but since degrees of longitude are farthest apart at the equator and converge at the poles, their distance varies greatly.

A degree of longitude is widest at the equator at 69.172 miles (111.321) and gradually shrinks to zero at the poles.

Therefore our buffer would have to be really large to account for the fact it shrinks to 0 at the poles

jupi3200: I think that dist2Degrees should produce a correct value.

When you try to figure out the distance between points on a sphere based on an angle, you need to to take the angle in the plane containing the sphere center and the two endpoints (not only latitude and longitude).

Using spherical approximation would provide the correct result with maximal 0.3% error (https://gis.stackexchange.com/questions/25494/how-accurate-is-approximating-the-earth-as-a-sphere)

lilian-hunt: It will produce the correct value, it's just that 1 degree = 0 km at the poles, so to check you would technically need a longitude_buffer =  ∞, rather than a latitude one = 0.00005. And there would be no improvement on the time. However there are probably not many locations at the poles and could pick some lower latitude as a cut off?

lilian-hunt: I could instead create my own private variable METERS_TO_LATITUDE_DEGREES_CONVERSION_FACTOR =  110.567km and then divide by DIST_BUFFER_METERS, instead of using these functions? If that's ok?

So it would be DIST_BUFFER_METERS / (METERS_TO_LATITUDE_DEGREES_CONVERSION_FACTOR * KILOMETER_TO_METER_CONVERSION_FACTOR)

jupi3200: 1 degree difference in the pole >> 0 km (as long as the vertex from which you calculate the rays is the center of earth).
1 latitude degree difference in the pole is 0 km.

Adding METERS_TO_LATITUDE_DEGREES_CONVERSION_FACTOR would require a good justification for the value. 

I am proposing using a LAT_BUFFER = dist2Degrees(DIST_BUFFER_METERS, EARTH_EQUATORIAL_RADIUS_KM  * KILOMETER_TO_METER_CONVERSION_FACTOR).
Can you please help me understand why using 0.00005 would be better?

lilian-hunt: Ok yes I understand now. I just realised I could use this function to convert. I initially was just looking at the DistanceUtils.calcLonDegreesAtLat() function and saw that they didn't have a calcLatDegreesAtLon() but using the EARTH_EQUATORIAL_RADIUS_KM we can calculate the conversion at the radius. I didn't quite understand how to use the function. But I have now implemented it as so.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/StopsTooCloseValidator.java
```diff
@@ -0,0 +1,98 @@
...
+
+  // If stops are closer than 5m based on distance calculated, generate a notice
+  static final int DIST_BUFFER_METERS = 5;
+
+  /**
+   * Convert DIST_BUFFER_METERS to degrees of latitude. Converts a distance in the units of the
+   * radius to degrees (360 degrees are in a circle). Assumming spherical earth model
+   */
+  private static final double LAT_BUFFER = DistanceUtils.dist2Degrees(DIST_BUFFER_METERS,
+      DistanceUtils.EARTH_EQUATORIAL_RADIUS_KM* KILOMETER_TO_METER_CONVERSION_FACTOR);
```
jupi3200: Nit:
DistanceUtils.EARTH_EQUATORIAL_RADIUS_KM * KILOMETER_TO_METER_CONVERSION_FACTOR)

lilian-hunt: Fixed. Interestingly clang-format-path . changes it from ..KM * KILOMETER.. to ..KM* KILOMETER.. without the first space between the star

jupi3200 (approved):

PR #31, opened 2021-02-08
Rule validates that trips in the same block have consistent route types

KathleenQ:
HashMap is used to store a block_id (as key) and its corresponding trip information (an AutoValue as value). The AutoValue includes trip's route type and other necessary information for notice generation. For a specific block_id, only the information of a trip that appears first for the block_id will be recorded. All other trips with the same block_id will be compared with the initially recorded trip. Their route types should be the same as the route type should not be changed inside a block; otherwise, a notice is generated.
HashMap is used to find the specific block_id more quickly in O(1) (as key). 
AutoValue is used to store other necessary trip fields as the value of HashMap, to extract specific field faster and easily. Compared to get an element in a List, the AutoValue field name is more straight-forward than the List index. In addition, by using AutoValue, you can know the specific AutoValue length rather than facing the unfixed length of extendable List.

### DIFF
--- a/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/BlockTripsWithInconsistentRouteTypesValidatorTest.java
+++ b/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/BlockTripsWithInconsistentRouteTypesValidatorTest.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright 2020 Google LLC, MobilityData IO
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


aannie-y (commented):
lgtm

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/BlockTripsWithInconsistentRouteTypesValidatorTest.java
```diff
@@ -0,0 +1,192 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsRouteTableContainer;
+import org.mobilitydata.gtfsvalidator.table.GtfsRouteType;
+import org.mobilitydata.gtfsvalidator.table.GtfsTrip;
+import org.mobilitydata.gtfsvalidator.table.GtfsTripTableContainer;
+
+@RunWith(JUnit4.class)
+public class BlockTripsWithInconsistentRouteTypesValidatorTest {
+  private BlockTripsWithInconsistentRouteTypesValidator validator =
+      new BlockTripsWithInconsistentRouteTypesValidator();
+  private final GtfsTrip trip11 =
```
jupi3200: Nit:
This test would be more readable if the trip names were more informative.

Perhaps something like _trip_bus_block_apple_

KathleenQ: Thank you for the advice! It really makes sense. I already changed trips' names.

jupi3200 (approved):

PR #32, opened 2021-02-08
Javascript logic

aannie-y:
Able to fetch response from Servlet and parse the JSON into the `callCorrespondingFunction` function which calls the corresponding function according to the error code. Also changed the Closure templating format to a table format to allow for better readability and added an additional template for `invalid_row_length`. Tests have also been implemented for for the `unknown_column` and `invalid_row_length` notices, notices not implemented and the `callCorrespondingFunction` to see if correct functions have been called.

Also removed a `response.write` line from RunValidatorServlet as it was not a JSON output and removed corresponding test lines from RunValidatorTest.java.


webapp/src/main/webapp/resources/output.js
```diff
@@ -12,18 +12,32 @@
...
+    // Notice has not been implemented, output raw json
+    if (runFunction(notice.code, notice) === -1) {
+      const unknownNotice = document.createElement('p');
+      unknownNotice.innerHTML = 'NOTICE CONTAINER NOT IMPLEMENTED - RAW JSON DATA: <br><br>' + JSON.stringify(notice);
+      document.getElementById("noticeContainer").appendChild(unknownNotice);
+    };
   }
+}
+
+function runFunction(name, arguments) {
```
jupi3200: perhaps a more readable name would be _runFunctionByName_

webapp/src/main/webapp/resources/output.js
```diff
@@ -12,18 +12,32 @@
...
+  const fn = window[name];
+  // check if fn is a function
+  if (typeof fn !== 'function') return -1;
```
jupi3200: If the function returns a value in some cases, if should probably heve well defined semantics and return a value in any case.

How about having this function return a boolean indicating if the notice was handled?

webapp/src/main/webapp/resources/output.js
```diff
@@ -12,18 +12,32 @@
...
-      }
-    ]
+function callCorrespondingFunction(noticesJSON) {
+  const noticeContainer = JSON.parse(noticesJSON);
+  for (var i = 0; i < noticeContainer.notices.length; i++) {
+    const notice = noticeContainer.notices[i];
+    // Notice has not been implemented, output raw json
+    if (runFunction(notice.code, notice) === -1) {
+      const unknownNotice = document.createElement('p');
+      unknownNotice.innerHTML = 'NOTICE CONTAINER NOT IMPLEMENTED - RAW JSON DATA: <br><br>' + JSON.stringify(notice);
```
jupi3200: Do we really want the prefix 'NOTICE CONTAINER NOT IMPLEMENTED - RAW JSON DATA:' repeated?

I would suggest that we have a list at the bottom of the page of notices not yet implemented, and adding the json to the list as needed.

webapp/src/main/webapp/resources/output.js
```diff
@@ -12,18 +12,32 @@
...
+  const fn = window[name];
+  // check if fn is a function
+  if (typeof fn !== 'function') return -1;
+  fn.apply(window, [arguments]);
+}
+
+function unknown_column(params) {
   const template = goog.soy.renderAsElement(validator.templates.unknownColumnNotice, params);
-  document.getElementById("warning").appendChild(template);
+  document.getElementById("noticeContainer").appendChild(template);
```
jupi3200: I thought we wanted one contained to warnings and another for errors.

webapp/src/main/webapp/script.js
```diff
@@ -11,3 +11,26 @@
...
+    }).then((resp) => {
+      // Check if output is in JSON format. If not, error has occured.
+      const contentType = resp.headers.get("content-type");
+      if (contentType && contentType.indexOf("application/json") !== -1) {
+        return resp.text().then(data => {
+          this.callCorrespondingFunction(data);
+        });
+      } else {
+        // Upload Error
+        this.alert(resp.text());
```
jupi3200: Do we really want a popup on the website?
How about just outputing the text in a red container?

webapp/src/main/webapp/test/output.test.js
```diff
@@ -18,9 +18,114 @@ describe('Output', function() {
...
+    const output = "<div><p class=\"error\">Error - Invalid csv row length!</p>\
+<p>Description: A row in the input file has a different number of values than specified by the CSV header.</p>\
+<p><b>1</b> Invalid row length found in:</p><table><thead><tr><th>Filename</th><th>CSV Row Number</th>\
+<th>Row length</th><th>Header count</th></tr></thead><tbody><tr><td>stop_times.txt</td>\
+<td>17</td><td>5</td><td>9</td></tr></tbody></table><p>Please set the row length as specified by the CSV header!</p>\
+<br><br></div><div><p class=\"warning\">Warning - Unknown Column(s) found!</p>\
+<p>Description: A column name is unknown.</p>\<p><b>1</b> unknown column(s) found in:</p>\
+<table><thead><tr><th>Filename</th><th>Field name</th><th>Index</th></tr></thead><tbody>\
+<tr><td>stop_times.txt</td><td>drop_off_time</td><td>8</td></tr>\
+</tbody></table><p>Please delete or rename column!</p><br><br></div>"
```
jupi3200: Could we use [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to make the test more readable?

jupi3200 (commented):
Can we possibly move the changes to UploadServlet and associated test to a different PR?

webapp/src/main/webapp/resources/output.js
aannie-y: yep ok

aannie-y: Done

webapp/src/main/webapp/resources/output.js
aannie-y: Hmm, yes ok. Will implement this.

webapp/src/main/webapp/test/output.test.js
aannie-y: I tried before but I get a syntax error for the backtick.

webapp/src/main/webapp/resources/output.js
aannie-y: Done!

webapp/src/main/webapp/resources/output.js
aannie-y: Ah yes, fixed.

webapp/src/main/webapp/script.js
aannie-y: Yep ok.

webapp/src/main/webapp/resources/output.js
jupi3200: Is there a reason we return an integer? Why not simple true/false ?

webapp/src/main/webapp/test/output.test.js
```diff
@@ -18,9 +19,116 @@ describe('Output', function() {
...
+        }
+      ]
+    };
+    invalid_row_length(params);
+    const output = "<div><p class=\"error\">Error - Invalid csv row length!</p>\
+<p>Description: A row in the input file has a different number of values than specified by the CSV header.</p>\
+<p><b>2</b> Invalid row length found in:</p><table><thead><tr><th>Filename</th><th>CSV Row Number</th>\
+<th>Row length</th><th>Header count</th></tr></thead><tbody><tr><td>stop_times.txt</td>\
+<td>17</td><td>5</td><td>9</td></tr><tr><td>stop_times.txt</td><td>18</td><td>5</td>\
+<td>9</td></tr></tbody></table><p>Please set the row length as specified by the CSV header!</p><br><br></div>"
```
jupi3200: Can you reformat this lines to make it more readable (Single line per row)?

webapp/src/main/webapp/resources/output.js
aannie-y: Fixed!

webapp/src/main/webapp/script.js
aannie-y: Fixed. Error now outputs in the DOM. CSS styling to be added later on to make it a red container.

webapp/src/main/webapp/test/output.test.js
aannie-y: Yep ok

aannie-y: Done. I've reformatted the lines so be a lot more readable.

webapp/src/main/webapp/resources/output.js
```diff
@@ -12,18 +12,37 @@
...
+  const noticeContainer = JSON.parse(noticesJSON);
+  for (var i = 0; i < noticeContainer.notices.length; i++) {
+    const notice = noticeContainer.notices[i];
+    // Notice has not been implemented, output raw json
+    if (!runFunctionName(notice.code, notice)) {
+      const unknownNotice = document.createElement('p');
+      unknownNotice.innerHTML = 'NOTICE CONTAINER NOT IMPLEMENTED - RAW JSON DATA: <br><br>' + JSON.stringify(notice);
+      document.getElementById("unimplementedNotices").appendChild(unknownNotice);
+    } else {
+      console.log("Notice has been handled!");
```
jupi3200: Can we avoid writing to console during non error execution?

webapp/src/main/webapp/resources/output.js
aannie-y: Fixed!

webapp/src/main/webapp/resources/output.js
aannie-y: Yep, will remove this.

aannie-y: Done!

jupi3200 (approved):

aannie-y:
> Can we possibly move the changes to UploadServlet and associated test to a different PR?

done

PR #33, opened 2021-02-09
Update uploadservlet.java and test to be compatible with template

aannie-y:
Removed unnecessary response.write() line which interfered with json parsing in the fetch.


KathleenQ (commented):
LGTM

jupi3200 (approved):

PR #43, opened 2021-02-11
CalendarHasNoActiveDays rule implemented

aannie-y:
Rule to check that the calendar has at least one active (available) day. Used a simple for loop and if statement to implement the rule. Loops through all the calendars in the given ArrayList of GtfsCalendar and checks whether the sum of the enum values from monday->sunday is less than 1. If it is less than 1, then all days are unavailable and a notice should be called.

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
@@ -20,18 +20,10 @@ import com.google.common.collect.ImmutableMap;
 import org.mobilitydata.gtfsvalidator.table.GtfsCalendarService;
 
 public class CalendarHasNoActiveDaysNotice extends Notice {
-  public CalendarHasNoActiveDaysNotice(String serviceId, GtfsCalendarService monday,
-      GtfsCalendarService tuesday, GtfsCalendarService wednesday, GtfsCalendarService thursday,
-      GtfsCalendarService friday, GtfsCalendarService saturday, GtfsCalendarService sunday) {
+  public CalendarHasNoActiveDaysNotice(String serviceId, long calendarCsvRowNumber) {
     super(new ImmutableMap.Builder<String, Object>()
               .put("serviceId", serviceId)
-              .put("monday", monday)
-              .put("tuesday", tuesday)
-              .put("wednesday", wednesday)
-              .put("thursday", thursday)
-              .put("friday", friday)
-              .put("saturday", saturday)
-              .put("sunday", sunday)
+              .put("calendarCsvRowNumber", calendarCsvRowNumber)
               .build());
   }
 
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.mobilitydata.gtfsvalidator.notice;
+
+import com.google.common.collect.ImmutableMap;
+import org.mobilitydata.gtfsvalidator.table.GtfsCalendarService;
+
+public class CalendarHasNoActiveDaysNotice extends Notice {
+  public CalendarHasNoActiveDaysNotice(String serviceId, GtfsCalendarService monday,
+      GtfsCalendarService tuesday, GtfsCalendarService wednesday, GtfsCalendarService thursday,
+      GtfsCalendarService friday, GtfsCalendarService saturday, GtfsCalendarService sunday) {
+    super(new ImmutableMap.Builder<String, Object>()
+              .put("serviceId", serviceId)
+              .put("monday", monday)
+              .put("tuesday", tuesday)
+              .put("wednesday", wednesday)
+              .put("thursday", thursday)
+              .put("friday", friday)
+              .put("saturday", saturday)
+              .put("sunday", sunday)
+              .build());
+  }
+
+  @Override
+  public String getCode() {
+    return "calendar_has_no_active_days";
+  }
+}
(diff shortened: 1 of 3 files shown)


gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidator.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public void validate(NoticeContainer noticeContainer) {
+    // Loop through calendar and add up enum values for each day. If value is less than 1, then no
+    // days are active.
+    for (GtfsCalendar calendar : calendarTable.getEntities()) {
+      if ((calendar.monday().getNumber() + calendar.tuesday().getNumber()
+              + calendar.wednesday().getNumber() + calendar.thursday().getNumber()
+              + calendar.friday().getNumber() + calendar.saturday().getNumber()
+              + calendar.sunday().getNumber())
+          < 1) {
```
jupi3200: Is there any guarantee that getNumber is positive for all enums?
Could there be an overflow?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
```diff
@@ -0,0 +1,42 @@
...
+      GtfsCalendarService friday, GtfsCalendarService saturday, GtfsCalendarService sunday) {
+    super(new ImmutableMap.Builder<String, Object>()
+              .put("serviceId", serviceId)
+              .put("monday", monday)
+              .put("tuesday", tuesday)
+              .put("wednesday", wednesday)
+              .put("thursday", thursday)
+              .put("friday", friday)
+              .put("saturday", saturday)
+              .put("sunday", sunday)
```
KathleenQ: I'm considering if you already know there must be no active day to generate the notice, you thus directly know all Monday to Sunday are NOT_AVAILABLE and do not need to put their 7 values into the notice. 
Instead, you can put some other fields in the notice, like csvRowNumber of the corresponding calendar, which is very useful.

jupi3200: Wouldn't all possible CalendarHasNoActiveDaysNotice have the same value for monday, tuesday, wednesday etc.?

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidatorTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsCalendar;
+import org.mobilitydata.gtfsvalidator.table.GtfsCalendarService;
+import org.mobilitydata.gtfsvalidator.table.GtfsCalendarTableContainer;
+
+@RunWith(JUnit4.class)
+public class CalendarHasNoActiveDaysValidatorTest {
+  private CalendarHasNoActiveDaysValidator validator = new CalendarHasNoActiveDaysValidator();
+
+  private final GtfsCalendar all_active_calendar =
+      new GtfsCalendar.Builder()
```
KathleenQ, line 38: csvRowNumber should be included.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
lilian-hunt: Good point! Every schema extends GtfsEntity which has csvRowNumber has the only attribute. So every table schema always stores the row number

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidator.java
aannie-y: can only get 0 or 1

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
aannie-y: yep I see. Will add this, thanks

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidator.java
jupi3200: But is this an implementation detail you should depend on?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
aannie-y: fixed

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidatorTest.java
aannie-y: yep fixed!

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/CalendarHasNoActiveDaysNotice.java
aannie-y: Yep, i have removed all that and added a csv row number instead.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidator.java
aannie-y: the enum values for calendar days are only 0 (for not available) and 1 (for available).

aannie-y: would just checking `calendar.monday().getNumber() === 'AVAILABLE'` for all the days be better then?

jupi3200: How about keeping a const variable for AVAILABLE number and then comparing each day with the specified number.

aannie-y: Fixed. Added variable NOT_AVAILABLE = 0 and used that to compare with the days.getNumber() instead.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/CalendarHasNoActiveDaysValidator.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mobilitydata.gtfsvalidator.table.GtfsCalendarTableContainer;
+
+/**
+ * Validates days in calendar.txt has at least one active (available) day.
+ *
+ * <p>Generated notices: {@link CalendarHasNoActiveDaysNotice}.
+ */
+@GtfsValidator
+public class CalendarHasNoActiveDaysValidator extends FileValidator {
+  private final static int NOT_AVAILBLE = 0;
```
jupi3200: Please don't trust current implementation details.

aannie-y: Fixed. Changed 0 to GtfsServiceCalendar.NOT_AVAILABLE.

jupi3200 (approved):

PR #46, opened 2021-02-11
Template for Stop too far from trip shape

KathleenQ:
Test is included.
The template looks like:
Screen Shot 2021-02-11 at 2 35 56 pm
The codes have already been run through clang-format.

### DIFF
--- a/webapp/src/main/webapp/resources/template.soy
+++ b/webapp/src/main/webapp/resources/template.soy
@@ -383,3 +383,38 @@
   <p>Please fix the stops positions to make their distance further than the trip buffer meters!</p>
   <br><br>
 {/template} 
+
+/**
+ * Renders a basic explanation of the stop too far from trip shape notice.
+ * @param notices : List<[stopId : string, stopSequence : number, tripId : string, shapeId : string, tripBufferMeters : number]>
+ */
+{template .stopTooFarFromTripShape}
+{let $numNotices: length($notices)/}
+  <p class="error">Error - Stop too far from trip shape!</p>
+  <p>Description: Stop is too far away from the trip shape.</p>
+  <p><b>{$numNotices}</b> Stop(s) too far from trip shape found in:</p>
+  <table>
+    <thead>
+      <tr>
+        <th>Stop ID</th>
+        <th>Stop Sequence</th>
+        <th>Trip ID</th>
+        <th>Shape ID</th>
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/resources/template.soy
```diff
@@ -382,4 +382,39 @@
...
+
+/**
+ * Renders a basic explanation of the stop too far from trip shape notice.
+ * @param notices : List<[stopId : string, stopSequence : number, tripId : string, shapeId : string, tripBufferMeters : number]>
+ */
+{template .stopTooFarFromTripShape}
+{let $numNotices: length($notices)/}
+  <p class="error">Error - Stop too far from trip shape!</p>
+  <p>Description: Stop is too far away from the trip shape.</p>
+  <p><b>{$numNotices}</b> Stop(s) too far from trip shape found in:</p>
```
lilian-hunt: This doesn't make sense grammatically. I also think there is too much repetition, so it is sufficient to say 
`<p><b>{$numNotices}</b> found:</p>`

aannie-y (commented):
besides what lilian said, lgtm

webapp/src/main/webapp/resources/template.soy
```diff
@@ -382,4 +382,39 @@
...
+          <td>{$notice.stopId}</td>
+          <td>{$notice.stopSequence}</td>
+          <td>{$notice.tripId}</td>
+          <td>{$notice.shapeId}</td>
+          <td>{$notice.tripBufferMeters}</td>
+        </tr>
+      {/foreach}
+    </tbody>
+  </table>
+  <p>Please fix the stop position to make it closer than the trip buffer meters from the trip shape!</p>
```
aannie-y: Bit confusing. Maybe word it like "Please fix the stop position to be within the trip buffer metres of the trip shape"

webapp/src/main/webapp/resources/template.soy
KathleenQ: Yes. Have adjusted all templates here.

webapp/src/main/webapp/resources/template.soy
KathleenQ: Yes. That sounds better! I have already adjusted it.

jupi3200 (approved):

PR #47, opened 2021-02-11
Create loading bar / result header

lilian-hunt:


webapp/src/main/webapp/script.js
```diff
@@ -17,6 +17,39 @@
...
+      errorOutput.parentNode.removeChild(errorOutput.nextSibling);
+    }
+    // Clear notice containers
+    errorOutput.innerHTML = '';
+    const error = document.getElementById('error');
+    error.innerHTML = '';
+    const warning = document.getElementById('warning');
+    warning.innerHTML = '';
+    const unimplementedNotices = document.getElementById('unimplementedNotices');
+    unimplementedNotices.innerHTML = '';
```
aannie-y: just wondering, why are you clearing the notice containers here?

aannie-y (commented):
lgtm

webapp/src/main/webapp/script.js
lilian-hunt: If you upload for a second time it will just keeping adding notices to the page because you are doing "appendChild" and there will be duplicates. Found this out through testing the webpage.

jupi3200: Can this function be split to make it more readable and avoid the need of comments to explain?

aannie-y: Ah nice

lilian-hunt: Yes now split into functions

jupi3200 (approved):

PR #55, opened 2021-02-12
Add template for Route unique names notice

KathleenQ:
I have changed the route type field in the notice from GtfsRouteType type to be the corresponding String type, to make it more convenient to use in the soy template.
Template looks like:
Screen Shot 2021-02-12 at 2 14 58 pm

### DIFF
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteUniqueNamesNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteUniqueNamesNotice.java
@@ -37,7 +37,7 @@ public class RouteUniqueNamesNotice extends Notice {
             .put("comparedRouteCsvRowNumber", comparedRouteCsvRowNumber)
             .put("routeLongName", routeLongName)
             .put("routeShortName", routeShortName)
-            .put("routeType", routeType)
+            .put("routeType", routeType.toString())
             .put("agencyId", agencyId)
             .build());
   }
--- a/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteUniqueNamesNotice.java
+++ b/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteUniqueNamesNotice.java
@@ -27,7 +27,7 @@ public class RouteUniqueNamesNotice extends Notice {
       long comparedRouteCsvRowNumber,
       String routeLongName,
       String routeShortName,
-      GtfsRouteType routeType,
+      String routeType,
       String agencyId) {
     super(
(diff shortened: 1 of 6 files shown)


gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/RouteUniqueNamesNotice.java
```diff
@@ -37,7 +37,7 @@ public RouteUniqueNamesNotice(
             .put("comparedRouteCsvRowNumber", comparedRouteCsvRowNumber)
             .put("routeLongName", routeLongName)
             .put("routeShortName", routeShortName)
-            .put("routeType", routeType)
+            .put("routeType", routeType.toString())
```
jupi3200: Is this a notice we created, or was it created by mobilitydata?

KathleenQ: It is a notice I created before.

aannie-y (commented):
LGTM

webapp/src/main/webapp/test/output.test.js
```diff
@@ -513,6 +513,53 @@ than one `agency_lang`, that\'s an error</p>\
...
+    const output =
+        '<div><p class="error">Error - Route without unique names!</p>\
+<p>Description: The combination of long name, short name, route type and agency ID for a route is not unique.</p>\
+<p><b>2</b> found:</p>\
+<table>\
+<thead>\
+<tr><th>Route ID</th><th>Route CSV Row Number</th><th>Compared Route ID</th><th>Compared Route CSV Row Number</th><th>Route Long Name</th><th>Route Short Name</th><th>Route Type</th><th>Agency ID</th></tr>\
+</thead>\
+<tbody>\
+<tr><td>routeC</td><td>8</td><td>routeF</td><td>2</td><td>RouteApple</td><td>apple</td><td>GtfsRouteType.SUBWAY</td><td>agency3</td></tr>\
```
jupi3200: GtfsRouteType.SUBWAY was never typed by any user. It's exposing implementation details that may suprise the user. Can we possibly avoid it?

KathleenQ: Yes. We can avoid it by using the corresponding number for GtfsRouteType instead, while the user may not know which GTFS Route Type the number corresponds to when reading the notice. Do you suggest in that way?

jupi3200: That would probably be even worse.
Can you think of a way to drop the 'GtfsRouteType.' prefix?

KathleenQ: Yes. I have used 'replace("GtfsRouteType.","")' to remove the prefix now.

jupi3200 (approved):

PR #57, opened 2021-02-12
Validator rule to check that transfers are unique

aannie-y:
Rule to check if transfers are unique. It uses a for loop, with time complexity O(n), to loop through the transfer list and an object, `FromToStop`, is created using Java's AutoValue with transfer's `from_stop_id` and `to_stop_id` values. A hash map, with `key: FromToStop` and `value: csvRowNumber,` is then used to store the `FromToStop` and its corresponding csv row number. A hash map was used instead of another array because the average time complexity of input and search is O(1). If another array was used, we would need to loop through the whole array to check if a value is unique and the time complexity for that would be O(n). Therefore using hash maps would have an overall time complexity of O(n) and using arrays would give us overall time complexity of O(n^2) and thus hash maps are more cost efficient.

An if statement then checks if the created fromToStop already exists in the hash map. If it does exist, it will create a notice with `from_stop_id`, `to_stop_id`, the existing `csvRowNumber` and the duplicated `csvRowNumber`. If the hash map does not contain the `fromToStop` object, it will be added into the hash map with its corresponding `csvRowNumber`.

### DIFF
--- a/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
+++ b/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
@@ -103,7 +103,7 @@ public class TransfersAreUniqueValidatorTest {
   }
 
   @Test
-  public void moreDuplicateTransfersShouldGenerateNotice() {
+  public void anotherDuplicateTransferShouldGenerateNotice() {
     final NoticeContainer noticeContainer = new NoticeContainer();
     List<GtfsTransfer> transfers =
         new ArrayList<>(Arrays.asList(transfer_1_duplicate, transfer_2, transfer_3, transfer_1));
--- a/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
+++ b/gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
@@ -19,6 +19,7 @@ package org.mobilitydata.gtfsvalidator.validator;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -62,7 +63,7 @@ public class TransfersAreUniqueValidatorTest {
   @Test
   public void allDifferentTransfersShouldNotGenerateNotice() {
     final NoticeContainer noticeContainer = new NoticeContainer();
-    List<GtfsTransfer> transfers = Arrays.asList(transfer_1, transfer_2, transfer_3);
+    List<GtfsTransfer> transfers = ImmutableList.of(transfer_1, transfer_2, transfer_3);
     validator.transferTable = GtfsTransferTableContainer.forEntities(transfers, noticeContainer);
     validator.validate(noticeContainer);
     assertThat(noticeContainer.getNotices()).isEmpty();
@@ -71,7 +72,7 @@ public class TransfersAreUniqueValidatorTest {
   @Test
   public void onlyDuplicateTransfersShouldGenerateNotice() {
     final NoticeContainer noticeContainer = new NoticeContainer();
-    List<GtfsTransfer> transfers = Arrays.asList(transfer_1, transfer_1_duplicate);
+    List<GtfsTransfer> transfers = ImmutableList.of(transfer_1, transfer_1_duplicate);
     validator.transferTable = GtfsTransferTableContainer.forEntities(transfers, noticeContainer);
     validator.validate(noticeContainer);
     assertThat(noticeContainer.getNotices())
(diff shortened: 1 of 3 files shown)


gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
```diff
@@ -0,0 +1,119 @@
...
+                             /* originalCsvRowNumber= */ 1),
+            new TransfersAreUniqueNotice(
+                /* fromStopId= */ "stop_3",
+                /* fromStopId= */ "stop_7",
+                /* duplicateCsvRowNumber= */ 4,
+                /* originalCsvRowNumber= */ 2));
+  }
+
+  @Test
+  public void moreDuplicateTransfersShouldGenerateNotice() {
```
KathleenQ: The name is a bit confusing for me - as it actually represents one duplicate transfer pair among some transfers, while I cannot get it from "more duplicate transfers". Maybe have a better name?

KathleenQ (commented):
Otherwise, LGTM.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
aannie-y: fixed! I changed the name to `anotherDuplicateTransferShouldGenerateNotice`. Name is trying to say that I have another test for a transfer. Not necessarily _another transfer_.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private final GtfsTransfer transfer_3 = new GtfsTransfer.Builder()
+                                              .setFromStopId("stop_5")
+                                              .setToStopId("stop_9")
+                                              .setCsvRowNumber(3)
+                                              .build();
+
+  @Test
+  public void allDifferentTransfersShouldNotGenerateNotice() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    List<GtfsTransfer> transfers =
```
tripd22: Is it necessary to convert the list into an ArrayList? Maybe give [this](https://www.baeldung.com/java-arrays-aslist-vs-new-arraylist) a read.

Can an immutable list be used?

aannie-y: Changed `new ArrayList<>(Arrays.asList())` to `Arrays.asList()`. When you said 'can an immutable list be used?' do you mean immutable as in cannot be structurally modified i.e. Arrays.asList() or do you want it fully immutable? I.e. using `Collections.unmodifiableList()` or guava's `ImmutableList<>()`?

jupi3200: I was suggesting using ImmutableList.of, but Arrays.asList is sufficient.

tripd22: Yeah, I was thinking [the Guava one](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/ImmutableList.html), i.e.

```
List<GtfsTransfer> transfers = ImmutableList.of(transfer_1, transfer_2, transfer_3);
```

I'm not sure how the `GtfsTransferTableContainer` treats the transfers so it might not work, but if it does it is probably better as immutable

aannie-y: Ok, i will check if it works with an ImmutableList.

aannie-y: Yep, ImmutableList works. I've changed all the `Arrays.asList` to `ImmutableList.of`.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
```diff
@@ -0,0 +1,120 @@
...
+                             /* duplicateCsvRowNumber= */ 10,
+                             /* originalCsvRowNumber= */ 1),
+            new TransfersAreUniqueNotice(
+                /* fromStopId= */ "stop_3",
+                /* fromStopId= */ "stop_7",
+                /* duplicateCsvRowNumber= */ 4,
+                /* originalCsvRowNumber= */ 2));
+  }
+
+  @Test
```
tripd22, line 111: One other thing you could test for is that reverse transfers don't trigger this notice. E.g. `stop_1 to stop_2` and `stop_2 to stop_1` shouldn't cause a notice. 

This would help prevent someone from breaking this behaviour with a future change to the implementation.

tripd22 (approved):

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/TransfersAreUniqueValidatorTest.java
aannie-y: done!

PR #58, opened 2021-02-12
Template for start and end time out of order

KathleenQ:
The template looks like:
Screen Shot 2021-02-12 at 5 26 56 pm
The codes have already been run through clang-format.

### DIFF
--- a/webapp/src/main/webapp/resources/template.soy
+++ b/webapp/src/main/webapp/resources/template.soy
@@ -552,3 +552,38 @@
   <p>Please delete the parent station of the station!</p>
   <br><br>
 {/template}
+
+/**
+ * Renders a basic explanation of the start and end time out of order notice.
+ * @param notices : List<[filename : string, csvRowNumber : number, entityId : string, startTime : string, endTime : string]>
+ */
+{template .startAndEndTimeOutOfOrder}
+{let $numNotices: length($notices)/}
+  <p class="error">Error - Start and end time out of order!</p>
+  <p>Description: start_time is after the end_time for a single entity.</p>
+  <p><b>{$numNotices}</b> found:</p>
+  <table>
+    <thead>
+      <tr>
+        <th>File Name</th>
+        <th>CSV Row Number</th>
+        <th>Entity ID</th>
+        <th>Start Time</th>
+        <th>End Time</th>
+      </tr>
+    </thead>
+    <tbody>
+      {foreach $notice in $notices}
+        <tr>
+          <td>{$notice.filename}</td>
+          <td>{$notice.csvRowNumber}</td>
+          <td>{$notice.entityId}</td>
+          <td>{$notice.startTime}</td>
+          <td>{$notice.endTime}</td>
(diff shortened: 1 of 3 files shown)


aannie-y (commented):
LGTM

webapp/src/main/webapp/resources/template.soy
```diff
@@ -552,3 +552,38 @@
...
 {/template}
+
+/**
+ * Renders a basic explanation of the start and end time out of order notice.
+ * @param notices : List<[filename : string, csvRowNumber : number, entityId : string, startTime : string, endTime : string]>
+ */
+{template .startAndEndTimeOutOfOrder}
+{let $numNotices: length($notices)/}
+  <p class="error">Error - Start and end time out of order!</p>
+  <p>Description: start_time is after the end_time for a single entity.</p>
```
jupi3200: Can we avoid the word 'entity' as the users might get confused by it. Would the word 'stop' be appropriate?

tripd22 (approved):
LGTM

webapp/src/main/webapp/resources/template.soy
KathleenQ: Sure. It actually is the start_time and end_time for a frequency. I will change "a single entity" to "a frequency" here.

KathleenQ: I have double checked it. I think it can be trip, stop etc, because the rule will be used on different files. So better not to specify it for only "stop" or "trip" here?

jupi3200: @tripd22 Can you please advise?

tripd22: I think this notice is probably only intended to be generated for [frequencies.txt](https://developers.google.com/transit/gtfs/reference#frequenciestxt), as it is the only file in the GTFS spec with start_time and end_time. [This](https://github.com/googleinterns/step268-2020/blob/dc98106db92cd00cddbcae0d406695eeb1a5a4a7/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/FrequencyTimeInOrderValidator.java#L43) is the only validation rule that generates it at the moment at least.

Given this, a better message might be `<p>Description: start_time is after the end_time for a row in frequencies.txt.</p>`.

It does seem a little strange that this notice has a `filename` in it if it was only for frequencies.txt, but there are already separate notices for stop_times - e.g. [departure time before arrival time](https://github.com/googleinterns/step268-2020/blob/master/gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/notice/StopTimeWithDepartureBeforeArrivalTimeNotice.java). This notice was added by some Mobility Data folk, so perhaps there was some miscommunication with them at some point?

KathleenQ: Sure. Good advice! I have changed that. Sorry that I missed the update for this notice in the new validator version.

jupi3200 (approved):

PR #63, opened 2021-02-12
template for route color contrast

aannie-y:
Add template for the routeColorContrastNotice.
Updated template with color box and string and collapsible button.
image

### DIFF
--- a/webapp/src/main/webapp/resources/template.soy
+++ b/webapp/src/main/webapp/resources/template.soy
@@ -559,29 +559,31 @@
  */
 {template .routeColorContrast}
 {let $numNotices: length($notices)/}
-  <p class="error">Error - Route Color Contrast!</p>
-  <p>Description: The contrast between the route and the route text is too similar!</p>
-  <p><b>{$numNotices}</b> found:</p>
-  <table>
-    <thead>
-      <tr>
-        <th>Route ID</th>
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/resources/template.soy
```diff
@@ -552,3 +552,36 @@
...
+        <th>Route Text Color</th>
+      </tr>
+    </thead>
+    <tbody>
+      {foreach $notice in $notices}
+        <tr>
+          <td>{$notice.routeId}</td>
+          <td>{$notice.csvRowNumber}</td>
+          <td>{$notice.routeColor}</td>
+          <td>{$notice.routeTextColor}</td>
```
jupi3200: Can we make this more visual - for example adding a box with the correct color?

aannie-y: What do you mean the 'correct color'? As in do you want me to put the colour of the route text and the route?

jupi3200: As discussed, show a filled squest with the specified color.

aannie-y: Done! See updated image as per comment.

aannie-y:
Updated look of the template (without collapsible CSS):
image

jupi3200 (approved):

PR #64, opened 2021-02-12
Template for route short and long name equal

aannie-y:
Template for route short and long name having equal name.

This is what the updated template looks like:
image

### DIFF
--- a/webapp/src/main/webapp/resources/template.soy
+++ b/webapp/src/main/webapp/resources/template.soy
@@ -559,27 +559,29 @@
  */
 {template .routeShortAndLongNameEqual}
 {let $numNotices: length($notices)/}
-  <p class="error">Error - Route Short and Long Name Equal!</p>
-  <p>Description: Both the short and long name of a route are equal.</p>
-  <p><b>{$numNotices}</b> found:</p>
-  <table>
-    <thead>
-      <tr>
-        <th>Route ID</th>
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/resources/template.soy
```diff
@@ -552,3 +552,36 @@
...
+  <p class="error">Error - Route Short and Long Name Equal!</p>
+  <p>Description: Both the short and long name of a route are equal.</p>
+  <p><b>{$numNotices}</b> found:</p>
+  <table>
+    <thead>
+      <tr>
+        <th>Route ID</th>
+        <th>CSV Row Number</th>
+        <th>Route Short Name</th>
+        <th>Route Long Name</th>
```
jupi3200: If they are equal, why do we need two columns?

aannie-y: True. I'll change it. Thanks

aannie-y: Done!

jupi3200: Can we update the screenshot?

jupi3200 (changes requested):
Can you please update the screenshot?

aannie-y:
> Can you please update the screenshot?

done

jupi3200 (approved):

PR #66, opened 2021-02-14
New rule for AmbiguousStopStationTransfers (with test)

KathleenQ:
This rule is to validate no ambiguous stop-to-station and station-to-stop transfers (i.e. if you have a stop-to-stop transfer, it will not have two (or more) stop-to-station or station-to-stop transfers corresponding to that).
All files have already been run through google-java-format for auto-formatting.

HashMap is used to store appeared stations' information for transfers. An AutoValue contains the corresponding from and to stations of a transfer and is used as the key of HashMap. The CSV row number of the corresponding transfer is the value. 
We need to store the CSV row number if a notice needs to be generated for the transfer and a new transfer, while the CSV row number should not be included in the comparison of stations in the AutoValue, thus we choose HashMap to store the extra information rather than using HashSet. 
Also, using HashMap means finding whether the key is contained and getting its value will take only O(1), which is time efficient.


KathleenQ (commented):
Ignore the soy template files (not for this PR).

jupi3200 (changes requested):
Please split the rule visualization (soy + javascript) from the rule implementation (Java)

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidator.java
```diff
@@ -0,0 +1,85 @@
...
+  public void validate(NoticeContainer noticeContainer) {
+    // Store all checked from and to stations of transfers and their corresponding csv row numbers
+    final Map<TransferIdentifier, Long> checkedStationsTransfer = new HashMap<>();
+    for (GtfsTransfer transfer : transferTable.getEntities()) {
+      final GtfsStop fromStop = stopTable.byStopId(transfer.fromStopId());
+      final GtfsStop toStop = stopTable.byStopId(transfer.toStopId());
+      // Only check station-to-stop or stop-to-station transfers
+      // If the sum of two locationTypes is 1, it must be 1 + 0 according to the locationType
+      // definition (i.e. a STOP and a STATION)
+      if (fromStop.locationType().getNumber() + toStop.locationType().getNumber() == 1) {
```
jupi3200: This depends essentially on enum mapping. Can we avoid it by explicitly comparing to the expected values?

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidator.java
```diff
@@ -0,0 +1,85 @@
...
+        // If the from_stop is a STOP with value 0, the transfer is from STOP to STATION; otherwise,
+        // the transfer is from STATION to STOP
+        final TransferIdentifier transferIdentifier =
+            (fromStop.locationType().getNumber() == 0)
```
jupi3200: Please don't depend on current enum mapping.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidatorTest.java
```diff
@@ -0,0 +1,223 @@
...
+          .setCsvRowNumber(7)
+          .setFromStopId("stationA")
+          .setToStopId("entrance2")
+          .build();
+
+  @Test
+  public void sameStationSameDirectionTransfersShouldGenerateNotice() {
+    final NoticeContainer noticeContainer = new NoticeContainer();
+    List<GtfsStop> stops =
+        new ArrayList<>(Arrays.asList(stationA, stationB, stop1_in_stationA, stop2_in_stationB));
```
jupi3200: As mentioned, please prefer Arrays.asList or (possibly better) use https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/ImmutableList.html#of-E-

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidatorTest.java
```diff
@@ -0,0 +1,223 @@
...
+    List<GtfsStop> stops =
+        new ArrayList<>(
+            Arrays.asList(
+                stationA,
+                stationB,
+                stationC,
+                stop1_in_stationA,
+                stop2_in_stationB,
+                stop3_in_stationC));
+    validator.stopTable = GtfsStopTableContainer.forEntities(stops, noticeContainer);
```
jupi3200: Why not just add all of the stops in a shared class constructor (as well as having a common noticeContainer)?

webapp/src/main/webapp/test/output.test.js
```diff
@@ -658,6 +658,117 @@ than one `agency_lang`, that\'s an error</p>\
...
+      notices: [
+        {
+          filename: 'FILENAME',
+          routeId: 'RouteB',
+          csvRowNumber: 12,
+          routeDesc: 'RouteFromAppleBlockToOrangeBlock',
+          specifiedField: 'route_short_name'
+        },
+        {
+          filename: 'FILENAME',
```
jupi3200: Do we expect variability in filename?

KathleenQ: Sorry, templates are not for this PR.

KathleenQ:
> Please split the rule visualization (soy + javascript) from the rule implementation (Java)

All templates are removed from this PR. They are already listed in other PRs.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidator.java
KathleenQ: Yes. Good suggestion! I have already fixed that.

gtfs-validator/main/src/main/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidator.java
KathleenQ: I have already changed the enum value comparisons to directly comparison using GTFS LocationType.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidatorTest.java
KathleenQ: This stops' combination is only used once in our tests - are we still preferred to create a shared class constructor for that?

jupi3200: If I read the tests correctly, each test adds some subest of stops to the stops table, and having more stops in the stops table would not have any impact. The logic of the tests is about the transfer table.
I am suggesting removing the common fixture from the test.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidatorTest.java
KathleenQ: Thank you for the advice! I have changed to use ImmutableList now.

gtfs-validator/main/src/test/java/org/mobilitydata/gtfsvalidator/validator/AmbiguousStopStationTransfersValidatorTest.java
KathleenQ: Good advice! I have made the shared stops and also a common noticeContainer in the latest version.

jupi3200 (approved):

PR #67, opened 2021-02-15
make errors/warnings collapsable

lilian-hunt:


lilian-hunt:
Only 2 lines have been changed but the HTML git diff is saying there is more
Screen Shot 2021-02-15 at 1 40 52 pm

Just changing the `<p>` tag to a button and adding some attributes and adding a div for the error content.

Steps:

`<button data-toggle="collapse" data-target="#[templateId]" class="[warning/error] collapsed">[Description]<span>+</span><p>-</p></button>`

change data-target to match template id with # in front `data-target="#[templateId]`
check that class is either `class="error collapsed"` or `class="warning collapsed"`
update the button text between `class = "error collapsed"> .... <span>`

add div
`<div class = "content collapse in" id="[templateId]">`

add closing div at the end 
`</div>`

webapp/src/main/webapp/style.css
```diff
@@ -27,6 +27,35 @@ td, th {
...
+  padding: 12px;
+}
+
+.error span, .warning span{
+  display: inline;
+}
+.error:not(.collapsed)  span, .warning:not(.collapsed)  span { 
+  display: none;
+}
+.error.collapsed p, .warning.collapsed p{
```
jupi3200, line 44: Should there be a line between this and the previous section?

jupi3200 (approved):

webapp/src/main/webapp/style.css
lilian-hunt: yes i will add one to be consistent

webapp/src/main/webapp/style.css
```diff
@@ -27,6 +27,37 @@ td, th {
 .warning {
   background: #ff7003;
   color: white;
+  
+}
+.content {
```
jupi3200, line 32: Can you please also add line between this and the previous section?

PR #78, opened 2021-02-17
Add visibility and headings to error and warning sections

KathleenQ:
I have set the error and warning sections to have visibility as well separately, and added headings to these two sections. Only when a warning is included in our generated notices, the "Warnings" heading will appear. Same for the "Errors" section.
I have already manually tested it. Corresponding JavaScript tests are also added.

### DIFF
--- a/webapp/src/main/webapp/index.html
+++ b/webapp/src/main/webapp/index.html
@@ -25,8 +25,8 @@
         <button class="btn btn-outline-primary" type="submit" id="uploadButton">Upload</button>
       </form>
       <div id="errorOutput"></div>
-      <div id="error"></div>
-      <div id="warning"></div>
+      <div id="error" style="visibility:hidden"></div>
+      <div id="warning" style="visibility:hidden"></div>
       <div id="unimplementedNotices"></div>
       <div id="allGood" style="visibility:hidden">
         <h4>All good!</h4>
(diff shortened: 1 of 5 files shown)


aannie-y (commented):
lgtm

lilian-hunt:
Please add screenshots of what this looks like: e.g. when there are errors and warnings: 
Screen Shot 2021-02-18 at 10 50 12 am
I personally think there should be some space between the errors and warnings style wise.

KathleenQ:
> Please add screenshots of what this looks like: e.g. when there are errors and warnings:
> Screen Shot 2021-02-18 at 10 50 12 am
> I personally think there should be some space between the errors and warnings style wise.

I have added space between error and warning sections and between warning sections and unimplemented notices.
Here are what it looks like now:
- For both error and warning sections:
Screen Shot 2021-02-18 at 11 06 27 am

- For only warning section (with unimplemented notices):
Screen Shot 2021-02-18 at 11 06 38 am

- For all good result:
Screen Shot 2021-02-18 at 11 06 53 am

For the 2nd and 3rd screenshots, for example, there are some space before "all good" section. That's because the titles (and the newly added space) for errors and warnings are hidden rather than no longer existing. So it is expected.

jupi3200 (approved):

